PR #1, opened 2021-05-07
LIIKUNTA-43 | Initial project

tuovinensanttu:
Adds the initial NextJs project with TypeScript.

To save a little bit of time `Dockerfile` and `docker-compose.yml` were copied from [palvelutarjotin-ui](https://github.com/City-of-Helsinki/palvelutarjotin-ui)

Added testing with Jest and testing-library.


.eslintrc.json
```diff
@@ -0,0 +1,36 @@
...
+        ],
+        "newlines-between": "always"
+      }
+    ],
+    "max-len": ["warn", { "code": 120 }],
+    "no-console": "warn",
+    "no-plusplus": "error",
+    "no-undef": "warn",
+    "object-curly-spacing": ["warn", "always"]
+  }
```
Jokinen, line 35: Are these configs copied from a project? Was it a NextJs project? Do they include react-hooks linters? I'd also consider adding the a11y linter: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y

.prettierrc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "bracketSpacing": true,
+  "tabWidth": 2,
+  "trailingComma": "es5",
+  "semi": true,
+  "singleQuote": true
```
Jokinen, line 6: Single quote is off by default, otherwise all of these re-declare default values. I'd just use the defaults from prettier directly and leave an empty object within this file, e.g. use double quotes and whatever else prettier recommends by default.

Jokinen (approved):
How was this bootstrapped? Did you use the helpers scripts? Just askin out of interest.

```
npx create-next-app
# or
yarn create next-app
```

.eslintrc.json
tuovinensanttu: I took these from `youth-membership` and checked against `palvelutarjotin-ui`. Seems there was not much difference between react and nextjs eslint files.

And to answer the above guestion I used `npx create-next-app`

.prettierrc.json
tuovinensanttu: I'll do that 👍

.eslintrc.json
tuovinensanttu: Added `"react/react-in-jsx-scope": "off",` and `    "globals": {
      "React": "writable"
    }`
These should disable the warning about import react to each component (which isn't required in NextJs)

PR #3, opened 2021-05-10
LIIKUNTA-31 | Add top navigation

Jokinen:
Adds top navigation to the application.

After this PR the application's landing page will render with a top navigation.

Some configurations were done to enable the navigation:
- HDS was added
- Helsinki Grotesk was added
- Layout logic was added
- Sass was added
- Meta data was added

In order to avoid implementing a too naive menu, I added mock data fetching logic for menu items.

This is sort of a problematic case with NextJS as currently there is no global, or app wide data fetching. See here: https://github.com/vercel/next.js/discussions/10949

Instead "global" data must be fetched view-by-view. This has the downside of creating duplicate code. I've implemented an initial approach for facilitating the fetching of menu data on each page.

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -1,0 +1,7 @@
+const path = require("path");
+
+module.exports = {
+  sassOptions: {
+    includePaths: [path.join(__dirname, "src/styles")],
+  },
+};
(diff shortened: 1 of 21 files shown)


tuovinensanttu (approved):
Tested changes locally with screen reader, everything is working nicely. Solid work 👍

PR #4, opened 2021-05-11
LIIKUNTA-26 | Allow user to select content language

Jokinen:
These changes toggle on the default i18n logic of NextJS. By default NextJS infers the user language, keeps track of the selected language, and synchronises the selected language with urls.

For now I did not see a reason to adopt a more full-fledged i18n library, like `i18next`. As we are sourcing our content from the CMS, translations will be managed on that end, and we only need to make sure that we are requesting content in the language the user expects. This should be easy enough to facilitate without a more encompassing library.

I'm still attempting to resolve how we can pass the current application language to the GraphQL endpoint in an elegant way.

In this PR I,
- Configure NextJS i18n feature
- Fetch supported languages from the API
- Integrate the language selector with NextJS's i18n feature
- Add SEO support for multiple languages

Note that `/` is the default language version (Finnish), and other languages versions are determined by their slug (for now `sv` and `en`).

There's a slight difference in how NextJS and the CMS perceive languages and locales. In NextJS, each language(-locale) is represented by a single string. In the CMS, each language is an object with multiple fields. The CMS has a `locale` and `slug` field for each language. The `slug` is meant to represent how the locale should appear in urls. NextJS does not offer have such fine grained control. Instead the locale strings configured in the next config are used as is in the url integration.

This means that the CMS is able to represent `en_US` with an url of `/en`. If we were to provide the language-locale string to NextJS, the url would be `/en_US`. In order to represent the language within the urls in the way the CMS data model guides, NextJS is configured in accordance with the slug field from the CMS data model. This can create a slight confusion, because it results in comparison between NextJS locales and CMS langauge.slugs.


.env.local.example
```diff
@@ -0,0 +1 @@
+NEXT_PUBLIC_CMS_GRAPHQL_ENDPOINT=https://liikunta.hkih.production.geniem.io/graphql
```
tuovinensanttu, line 1: Missing new line.

.env.test
```diff
@@ -0,0 +1 @@
+NEXT_PUBLIC_CMS_GRAPHQL_ENDPOINT=https://liikunta.hkih.production.geniem.io/graphql
```
tuovinensanttu, line 1: Same here.

tuovinensanttu (approved):
Nice work!

PR #6, opened 2021-05-12
LIIKUNTA-58 | Add popular now section

Jokinen:
Adds a mock "popular no" section.

Takes inspiration and copies code from Tapahtumat.

I used an approach with the cards where they are not wrapped in an interactive element such as a link to make them easier to use with screen readers. However, I implemented some special logic for mouse users which allows them to click anywhere on the card to fire the navigation routine.

I also implemented the "call to action" (the arrow button) in a way where it's not focusable and as such does not introduce extra focusable elements, but does still communicate it's presence to sighted screen reader users (niche uscases).

I adopted a MVVM pattern with the card for now. I'll develop the approach further while implementing collections.

I could not find recommendations from WP so I was not able to mimic the data model.

### DIFF
--- a/src/components/card/Card.tsx
+++ b/src/components/card/Card.tsx
@@ -1,0 +1,99 @@
+import React, { useRef } from "react";
+import Link from "next/link";
+import { IconArrowRight } from "hds-react";
+
+import { Item } from "../../types";
+import Text from "../text/Text";
+import Keyword from "../keyword/Keyword";
+import styles from "./card.module.scss";
+
+function Card({ id, title, infoLines, keywords, pre, href, image }: Item) {
+  const linkRef = useRef(null);
+  const downRef = useRef<Date>(null);
+
+  const handleWrapperMouseDown = (e: React.MouseEvent<HTMLElement>) => {
+    downRef.current = new Date();
+  };
+
+  const handleWrapperMouseUp = (e: React.MouseEvent<HTMLElement>) => {
+    const link = linkRef.current;
+    const up = new Date();
+
+    // Invoke a click on link if
+    // 1. The event did not bubble from the link itself
+    // 2. The user is not attempting to select text
+    if (link !== e.target && up.getTime() - downRef.current.getTime() < 200) {
+      link.click();
+    }
+  };
+
+  return (
+    // These listeners allow us to provide a better UX for mouse while not
+    // hampering the UX for keybord or screen reader users.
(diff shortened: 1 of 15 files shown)


src/components/card/Card.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  return (
+    // These listeners allow us to provide a better UX for mouse while not
+    // hampering the UX for keybord or screen reader users.
+    // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions
+    <article
+      onMouseDown={handleWrapperMouseDown}
+      onMouseUp={handleWrapperMouseUp}
+      className={styles.card}
+    >
```
tuovinensanttu, line 38: As the whole `acticle` is considered a link, should it have hover effect for the cursor? (This might just be me, but it feels weird hovering on a big card like this without cursor changing 😅 )

src/components/layout/layout.module.scss
```diff
@@ -23,4 +26,15 @@
...
+  .main {
+    grid-column: 1 / -1;
+
+    display: inherit;
+    grid-template-columns: inherit;
+
+    & > * {
+      grid-column: 2;
+    }
+  }
```
tuovinensanttu, line 39: Thanks for this! Now I don't have to figure out how to span the Hero section on all three columns 😄

tuovinensanttu (approved):
Damn this is looking smooth!

src/components/keyword/Keyword.tsx
```diff
@@ -0,0 +1,40 @@
...
+    <button
+      className={classNames(styles.keyword, color && styles[color], {
+        [styles.blackOnMobile]: blackOnMobile,
+        [styles.hideOnMobile]: hideOnMobile,
+      })}
+      onClick={handleClick}
+      type="button"
+    >
+      {keyword}
+    </button>
```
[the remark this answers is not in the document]
Jokinen: Yes, but they should lead to a 404 page.

src/components/card/Card.tsx
Jokinen: You are correct, I'll add this.

src/components/keyword/Keyword.tsx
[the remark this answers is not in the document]
tuovinensanttu: Tested this again, indeed the url leads to `/keywords` and clicking the whole card forwards to `/event` 👍

PR #7, opened 2021-05-14
LIIKUNTA-57 | Add collections section to landing page

Jokinen:
* Modifies the card component so that other card variants can be created with them
* Creates a collection grid variant in the list component
* Lists collections on the landing page


src/components/card/Card.tsx
```diff
@@ -1,13 +1,175 @@
...
+}
+
+type CardKeywordsProps = {
+  keywords: KeywordType[];
+  className?: string;
+};
+
+// This is a visual only guide that tells visual users that this card
+// is clickable. It has a label so that visual screen reader users
+// can interact with the element.
```
Jokinen, line 124: Comment that was copied by accident

src/components/card/Card.tsx
```diff
@@ -1,13 +1,175 @@
...
+  );
+}
+
+type CardImageProps = {
+  image: string;
+};
+
+// This is a visual only guide that tells visual users that this card
+// is clickable. It has a label so that visual screen reader users
+// can interact with the element.
```
Jokinen, line 149: Comment that was copied by accident

src/api/LiikuntaApolloClient.ts
```diff
@@ -0,0 +1,108 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment, @typescript-eslint/no-explicit-any */
```
Jokinen, line 1: Apollo's types are not accurate which is why type checking results have been ignored in this file.

src/api/LiikuntaApolloClient.ts
```diff
@@ -0,0 +1,108 @@
...
+
+type WithPageGlobals<T> = T & {
+  global: {
+    language: Language[];
+    menuItems: MenuItem[];
+  };
+};
+
+class LiikuntaApolloClient extends ApolloClient<NormalizedCacheObject> {
+  async pageQuery<T = any, TVariables = OperationVariables>(
```
Jokinen, line 53: Allows pages to require global data without much duplicate markup. Until `getStaticProps` is available in `_app` this approach allows data that's required on each page to be shared (menu, language and so forth).

src/api/apolloCmsClient.ts
```diff
@@ -1,8 +1,9 @@
-import { ApolloClient, InMemoryCache } from "@apollo/client";
+import { InMemoryCache } from "@apollo/client";
 
 import Config from "../config";
+import LiikuntaApolloClient from "./LiikuntaApolloClient";
 
-const cmsClient = new ApolloClient({
+const cmsClient = new LiikuntaApolloClient({
```
Jokinen, line 6: This client should be configured with `ssr: true`

src/config.ts
```diff
@@ -1,18 +1,14 @@
...
+      throw Error(`Environment variable with name ${variable} was not found`);
     }
 
-    return variableValue;
+    return variable;
   }
 
   static get cmsGraphqlEndpoint() {
-    return Config.getEnvOrError("NEXT_PUBLIC_CMS_GRAPHQL_ENDPOINT");
+    return Config.getEnvOrError(process.env.NEXT_PUBLIC_CMS_GRAPHQL_ENDPOINT);
```
Jokinen, line 11: After this PR, the Apollo client will also be created in the client. Within the client environment, environment variables can be accessed only by explicitly declaring the field you want (`process.env.NEXT_PUBLIC_CMS_GRAPHQL_ENDPOINT`) and dynamic requests (`process.env[fieldName]`) are not possible.

src/api/LiikuntaApolloClient.ts
```diff
@@ -0,0 +1,106 @@
...
+
+const GLOBAL_QUERY = gql`
+  {
+    globalLanguages: languages {
+      id
+      name
+      slug
+      code
+      locale
+    }
```
tuovinensanttu, line 36: I think in all the other front-end applications the graphql queries & mutations are in their own files. Do we wan't to make same structure here or keep them as they are now?

tuovinensanttu (approved):
The apollo client creation was pretty technical, but the link provided in the comments opened up the logic a little bit more. 

The changes to Card component is really nice and allows re-usability.

src/api/LiikuntaApolloClient.ts
Jokinen: I think that this particular query is is more of a temporary hack than something that should remain for the long term--in that sense I think that this should be managed within this file so that it's easy to remove later.

As for the other queries, I  don't have as clear of an opinion. Apollo recommends storing queries alongside components. For the page components, the only place is the file itself. For other components, it's possible to store the queries alongside the component file in a separate file.

I don't really have a strong opinion. By default I would maybe prefer the more simple approach which I my opinion is to not create a separate file What pros and cons do you see? Do you prefer one of the approaches?

Jokinen: I'll merge this PR, but we can continue the discussion. Let's make another PR to fix this if we decide to. I have other changes that are awaiting that these commits land on main.

PR #11, opened 2021-05-19
LIIKUNTA-29 | Fetch page data in a separate query

Jokinen:
In preparation for LIIKUNTA-29, simplify `pageQuery` logic and declare `RootQuery` as a query type in order to allow Apollo's cache to be built correctly.


tuovinensanttu (approved):
👍

PR #13, opened 2021-05-25
LIIKUNTA-61 | Unified search integration

tuomas777:
Implemented initial search functionality using the [Unified Search.](https://github.com/City-of-Helsinki/unified-search)

There is currently only free text search for locations available, and the amount of information provided about those is also very limited.

The layout for displaying search results is also very initial and just there to get some visible results, and will need lots of love once we know what kind of info exactly needs to be there.

### DIFF
--- a/src/pages/search/index.tsx
+++ b/src/pages/search/index.tsx
@@ -1,12 +1,82 @@
 import { GetStaticPropsContext } from "next";
+import { useRouter } from "next/router";
+import { gql, useQuery } from "@apollo/client";
+import React from "react";
 
-import Page from "../../components/page/Page";
 import SearchPageSearchForm from "../../components/search/searchPageSearchForm/SearchPageSearchForm";
+import Page from "../../components/page/Page";
+import searchApolloClient from "../../api/searchApolloClient";
+import { Connection, Item, Keyword, SearchResult } from "../../types";
+import { getNodes } from "../../api/utils";
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const noop = () => {};
+
+export const SEARCH_QUERY = gql`
+  query SearchQuery($q: String) {
+    unifiedSearch(q: $q, index: "location", ontology: "Liikunta") {
+      edges {
+        node {
+          venue {
+            name {
+              fi
+            }
+            description {
+              fi
+            }
+          }
+        }
+      }
+    }
+  }
+`;
+
+const emptyConnection = {
+  edges: [],
+};
+
+const mockKeywords: Keyword[] = [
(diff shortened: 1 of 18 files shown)


src/pages/search/search.module.scss
```diff
@@ -0,0 +1,30 @@
+@import "variables";
+
+.section {
```
Jokinen, line 3: Do you use this section class somewhere?

src/pages/search/index.tsx
```diff
@@ -1,12 +1,101 @@
...
       <SearchPageSearchForm />
+      <Koros className={styles.koros} />
+      {loading ? (
+        <LoadingSpinner className={styles.spinner} />
+      ) : (
+        <Section variant="contained">
+          <Text
+            variant="h2"
+            className={styles.resultCount}
+          >{`${searchResultItems.length} hakutulosta`}</Text>
```
Jokinen, line 90: Perhaps you should include a `role="status"` here

https://www.w3.org/WAI/WCAG21/working-examples/aria-role-status-searchresults/

I have a hunch that adding the attribute alone may not be enough. For instance, the result count may not get announced if the result count doesn't change. On the other hand, it's possible that the attribute works poorly with React and that you would need to consider some sort of aria-live package to be able to add proper support. This is a bit of a trial and error task where you would need to use Safari and VoiceOver to find out what works well enough.

Does TypeScript complain if you do:
```suggestion
          >{searchResultItems.length} hakutulosta</Text>
```

src/pages/search/index.tsx
```diff
@@ -1,12 +1,101 @@
...
+          <List
+            variant="searchResult"
+            items={searchResultItems.map((item, index) => (
+              <SearchResultCard key={index} {...item} />
```
Jokinen, line 94: Using `index` as the key not encouraged. It should only be used when there's no other feasible approach. As far as I can see, you are generating and `id` for search results. That would be a better option.

React expects that they `key` can be used to identify an element between renders.

Jokinen (approved):
There are a few layout issues in Safari, but my guess is that they have existed beforehand.

The images get quite big:

Screenshot 2021-05-25 at 9 24 18

On very small screens the content of the page overflows the screen, because the keywords' flex container doesn't allow wrapping.

All in all these changes look very good, good job 👍

src/pages/search/search.module.scss
tuomas777: Wow, well spotted 🙂

src/pages/search/index.tsx
tuomas777: Ah of course, silly me.

PR #14, opened 2021-05-26
LIIKUNTA-63 | Add suggestion to free text search

Jokinen:
Now while the user types, the UI should offer suggestions for the user. The suggestions should be in the currently chosen language.

### Custom Dropdown implementation with Downshift

The Combobox in HDS doesn't support our usecase. The combobox is designed for selecting one option from many. Our usecase is a text field that allows the user to optionally choose from a curated option list. The technical detail that makes it impossible to use Combobox is that it doesn't allow us to control the value of the input, which is why we are not able to track the value of the input--and hence make searches with it.

I implemented a custom dropdown which uses the same tooling as HDS does for its dropdown implementation.

The implementation lends more from HDS than from the implementation in events.

### Debouncing

I added a very basic debouncing measure. This should limit the amount of times the UI reacts to suggestion request responses. In other words, if users write many letters in quick succession, only the search done after the final letter will get rendered in the UI. I'm not sure but the request themselves may also get debounced.

This isn't an entirely foolproof approach with Apollo, but it seemed to work with this usecase.

### DIFF
--- a/src/api/utils.ts
+++ b/src/api/utils.ts
@@ -21,3 +21,10 @@ export function getQlLanguage(language: string) {
     en: "EN",
   }[language];
 }
+export function getUnifiedSearchLanguage(language: string) {
+  return {
+    fi: "FINNISH",
+    sv: "SWEDISH",
+    en: "ENGLISH",
+  }[language];
+}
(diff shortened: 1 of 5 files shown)


tuovinensanttu (approved):
Looks good to me!

PR #15, opened 2021-05-28
LIIKUNTA-62 | Search page pagination

tuovinensanttu:
Add pagination to `SearchPage`. 

To get `fetchMore` results to frontend I needed to add `merge` function to apollo's cache (this was suggested by [documentation](https://www.apollographql.com/docs/react/pagination/core-api/#merging-paginated-results)). This however broke the functionality of triggering query again by simply using params we got from the URL. I solved this by manually toggling refetch when search - button is pressed.

List should now be accessible, when more results are loaded we store focus on extra `li` element, which also tells SR users how many new results there are. I decided to hide this from visual users.


tuomas777 (commented):
Hmm, I might be missing something, but doesn't this implementation need to fetch all of the preceding results again also every time "load more" is pressed? It seems that there will potentially be a lot of unnecessary data fetched, and some huge queries at the end 🙂

Also, did you try using Apollo's `relayStylePagination` helper?

tuovinensanttu:
> Hmm, I might be missing something, but doesn't this implementation need to fetch all of the preceding results again also every time "load more" is pressed? It seems that there will potentially be a lot of unnecessary data fetched, and some huge queries at the end 🙂
> 
> Also, did you try using Apollo's `relayStylePagination` helper?

Nope, you are not missing anything, seems that I can't read documentation properly and totally missed that I could use endCursor as variable for the next search. With that change I also replaced the custom merge function with `relayStylePagination `

src/components/list/SearchList.tsx
```diff
@@ -0,0 +1,70 @@
...
+  ) => {
+    const resultsLeft = count ? count - items.length : 0;
+    const a11yIndex = (Math.floor(items.length / blockSize) - 1) * blockSize;
+    const loadedMoreAmount = hasNext ? blockSize : count - items.length;
+
+    return (
+      <>
+        {!loading && (
+          <Text variant="h2" className={styles.resultCount} role="status">
+            {items.length} hakutulosta
```
tuomas777, line 30: I think we would rather want to show the total result count here?

tuovinensanttu: That would indeed make more sense.

tuovinensanttu:
After UnifiedSearch fixes pagination seems to be working.

tuomas777 (approved):
Looking great AFAICT 👍 🚀

PR #16, opened 2021-05-31
Chore | Increase layout margin on desktop width.

tuovinensanttu:
Increased margin used on desktop width.

### DIFF
--- a/src/styles/breakpoints.scss
+++ b/src/styles/breakpoints.scss
@@ -7,7 +7,7 @@ $breakpoints: (
   l: $breakpoint-l,
   xl: $breakpoint-xl,
   // xxl => 1568px
-  xll: $breakpoint-xl + $breakpoint-xs,
+  xxl: $breakpoint-xl + $breakpoint-xs,
 );
 
 @mixin respond-above($breakpoint) {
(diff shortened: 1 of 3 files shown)


Jokinen (commented):
Collections, categories and recommendations don't seem to behave as nicely with this spacing when using a laptop screen. Perhaps the `List` component should be adjusted to take the new spacing into account? By experimenting with the `--base-list-item-width` variable, you can find which widths work well with this screen size.

Screenshot 2021-06-01 at 13 29 01

Tbh, I think that the behaviour might've been bad before as well, because I only checked with my laptop screen size. 1366 x 768 size should be a good standard for a laptop screen.

The mobile view is the most important by far, but the laptop view is perhaps among the distant seconds.

Jokinen (approved):

src/styles/breakpoints.scss
```diff
@@ -6,6 +6,8 @@ $breakpoints: (
   m: $breakpoint-m,
   l: $breakpoint-l,
   xl: $breakpoint-xl,
+  // xxl => 1568px
+  xll: $breakpoint-xl + $breakpoint-xs,
```
tuomas777, line 10: I guess this was meant to be `xxl` ? Also, both forms `xll` and `xxl` seem to be used in the current changes 🙂

tuovinensanttu: Good catch! I'll fix these.

tuomas777 (approved):
:shipit:

PR #17, opened 2021-05-31
LIIKUNTA-65 | Redirect front page search submit to search page

tuomas777:


src/components/search/landingPageSearchForm/LandingPageSearchForm.tsx
```diff
@@ -21,6 +38,8 @@ function LandingPageSearchForm() {
         id="q"
         label="Mitä etsit?"
         placeholder="Aloita kirjoittamalla tähän, esim. uimahalli tai jooga"
+        onChange={handleChange}
+        value={searchText}
```
Jokinen, line 42: I wouldn't make this a controlled component because the search form on the landing page will always be very simple and "one way bindings" should be enough for it. Instead I'd just use it as uncontrolled. It simplifies the code a bit (e.g. you would use the submit event in the onSubmit handler to find form values) and gives a very minor performance boost.

But I would not also spend the time changing this to an uncontrolled component. This approach is perfectly fine and would probably be considered a best practice by some.

Jokinen (approved):

src/components/search/landingPageSearchForm/LandingPageSearchForm.tsx
tuomas777: I actually pondered about that a bit, and then decided to go with a controlled one because I happened to read someone's opinion somewhere that you should practically never use uncontrolled forms 🙂 Could have made sense here.

I have a feeling we might need to modify this in the future anyway, as Tapahtumat has autocomplete on the front page as well.

PR #19, opened 2021-06-01
Chore | Adjust queries so that they are not brittle in production use

Jokinen:
In this PR I make the following adjustments to queries:

1. I alter the landing page query so that it's made according to the best practices that Geniem advices
2. I change queries so that they are language aware instead of using a hardcoded Finnish value. This means that missing translations will be missing in the application. Translations can be added through the CMS.
3. I added application menus to the CMS (translated) and removed the mock data. I added a `rewrites` config which shows how we can use translated url structures for a better user experience. I applied translations for the search view.
4. I added a page query into the search view. The search view was missing the logic that initializes the application state with "global" data, or maybe better described "layout data". This means things like menus and languages.

As an adjustment I added logic that persists the user's language choice for the remaining session. This way language autodetection doesn't overrule user choice.

I wasn't able to adjust the collection query, because it's still unclear what steps we should take to circumvent an ongoing issue with it.


next.config.js
```diff
@@ -10,4 +10,16 @@ module.exports = {
...
+  async rewrites() {
+    return [
+      {
+        source: "/haku",
+        destination: "/search",
+      },
+      {
+        source: "/sok",
+        destination: "/search",
+      },
```
tuovinensanttu, line 22: If I understand this correctly this is swedish translation for search? If so shouldn't it be `sök`?

tuovinensanttu (approved):
As far as I can tell these changes are looking good!

next.config.js
Jokinen: Cleared this up on Slack 👍

PR #21, opened 2021-06-02
LIIKUNTA-64 | Fix FOUC with statically optimized pages

Jokinen:
I tried my hardest to avoid creating a custom server. A custom server would remove some Next features and optimizations.

I was able to find this approach where we create a fake document and then attempt to reset it between requests.

This is a bandaid fix in that the document is not created per request. I wasn't able to find a better way to hook into the application logic early enough.

The ideal fix would be to add SSR support into HDS.


tuovinensanttu (approved):
Nice job 👍

PR #24, opened 2021-06-08
LIIKUNTA-77 | Add api for sport venue details

tuovinensanttu:
Initial API for fetching data for sport venue details page. For now there's only two possible data sources, tprek & linkedevents.


src/pages/api/venues/[venueId].ts
```diff
@@ -0,0 +1,20 @@
...
+
+import formatResponseObject from "../../../util/api/formatResponseObject";
+import getSourceUrl from "../../../util/api/getSourceUrl";
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse
+) {
+  const { venueId } = req.query;
+  const splitVenueId = (venueId as string).split(":");
```
Comment (Jokinen) [inline src/pages/api/venues/[venueId].ts:12]: Option for code style:

```suggestion
  const [source, id] = (venueId as string).split(":");
```

src/types.ts
```diff
@@ -83,3 +83,28 @@ export type Venue = {
...
+
+export type TranslationsObject = {
+  fi: string;
+  en: string;
+  sv: string;
+};
+
+export type VenueDetails = {
+  id: string;
+  data_source: string | null;
```
Jokinen, line 95: Should we use `camelCase` considering we are within a JS app and are creating an API that will only be used by a JS based app?

src/util/api/formatResponseObject.ts
```diff
@@ -0,0 +1,65 @@
+import { camelCase, mapKeys, pick, rearg } from "lodash";
```
Jokinen, line 1: About the location of this helper.

We also have an API folder in `src` root that has an util file. I don't know what we would do to improve it, but we should somehow clear up the responsibilities between `/ util/api` from `/api/utils`. There are multiple levels of conflict.

Perhaps the should rename `src/api` to something else , like `src/client` 🤔

src/pages/api/venues/[venueId].ts
```diff
@@ -0,0 +1,27 @@
...
+
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse
+) {
+  const { venueId } = req.query;
+  const [source, id] = (venueId as string).split(":");
+
+  if (!source || !id) {
+    res.status(400).json({ msg: "Invalid ID parameter" });
```
Comment (Jokinen) [inline src/pages/api/venues/[venueId].ts:15]: Wondering why we are using `msg` instead for instance `message`?

src/util/api/formatResponseObject.ts
```diff
@@ -0,0 +1,65 @@
...
+          ...(obj.address_city_fi && { fi: obj.address_city_fi }),
+          ...(obj.address_city_en && { en: obj.address_city_en }),
+          ...(obj.address_city_sv && { sv: obj.address_city_sv }),
+        },
+        contactType: null,
+        infoUrl: null,
+        telephone: null,
+      } as VenueDetails;
+    default:
+      return obj;
```
Jokinen, line 63: By default shouldn't this error? The resulting object is unlikely to be correct format if there's no parser for it.

src/util/api/formatResponseObject.ts
```diff
@@ -0,0 +1,65 @@
...
+        "postal_code",
+        "image",
+        "address_locality",
+        "position",
+        "description",
+        "name",
+        "info_url",
+        "street_address",
+        "telephone",
+      ]);
```
Jokinen, line 29: As a question, does linked's data model just happen to align well with the intended data model for `VenueDetails` or does `VenueDetails` copy the data model of linked?

src/util/api/getSourceUrl.ts
```diff
@@ -0,0 +1,11 @@
+export default function getSourceUrl(source: string, id: string) {
+  switch (source) {
+    case "linked":
+      const linkedId = `tprek:${id}`;
+      return `https://api.hel.fi/linkedevents/v1/place/${linkedId}/`;
+    case "tprek":
+      return `https://www.hel.fi/palvelukarttaws/rest/v4/unit/${id}/`;
+    default:
+      return `https://www.hel.fi/palvelukarttaws/rest/v4/unit/${id}/`;
```
Jokinen, line 9: Do we want to default to a value instead of erroring?

Jokinen (approved):

src/util/api/formatResponseObject.ts
tuovinensanttu: Totally forgot that there was an api folder under src. I'll rename that one to `client` as you suggested.

src/pages/api/venues/[venueId].ts
Comment (tuovinensanttu) [reply src/pages/api/venues/[venueId].ts:15]: This is just a habit of mine that I have used on my own projects. I'll change these to `message`.

src/util/api/formatResponseObject.ts
tuovinensanttu: Thats true, I need to think this over.

src/util/api/formatResponseObject.ts
tuovinensanttu: I took a look at responses I got from each url and decided to use linked's response as a base. Imo it made a little more sense.

Jokinen: Ok. Personally I would have thought about the data model in terms of the requirements of the UI and use neither data source's model as the base. If we were to use a base, then perhaps unified search could have been a candidate, because sharing the model of the data between the search and detail views could have provided some benefits.

This is a minor detail, but as it stands now, it seems that this approach makes it likely that future developers will also copy linked's data model as is--as opposed to approaching the data model through the requirements of the UI.

Of course, linked is a well crafted service and the data model it provides is likely perfectly fine for us. As far as I can see, there's nothing I would want to change outright.

tuovinensanttu: 😄  don't get me wrong, of course I gave this a thought on UI's standpoint as well. Even if I'd start all over again `VenueDetails` is like 99% something that I would have returned anyway.

.github/workflows/production.yml
```diff
@@ -21,8 +21,8 @@ env:
   K8S_LIMIT_CPU: 400m
   K8S_LIMIT_RAM: 400Mi
   K8S_REPLICACOUNT: 2
-  K8S_LIVENESS_PATH: /api/healthz
-  K8S_READINESS_PATH: /api/readiness
+  K8S_LIVENESS_PATH: /client/healthz
```
Jokinen, line 24: This is probably wrong?

tuovinensanttu: Yup, seems that my IDE derped a ton with this rename. First the functianility would't go through and now it seems it fixed automatically all the wrong files and didn't touch the ones that actually needed fixing.

PR #25, opened 2021-06-08
Liikunta 71 | Add details page with hardcoded content

Jokinen:
Adds a preliminary version of the details page.

The details page can be accessed at the path `/ entity/[id]`.

As a tangentially related change I fixed the logic for changing language.

Screenshot 2021-06-08 at 11 21 19
Screenshot 2021-06-08 at 11 21 41


tuovinensanttu (approved):
Looks good to me. I was just wondering should there be one more breakpoint (s / xs). When the layout changes with the current breakpoint there is a lot of space on the right side (aside section with blocks). As the blocks are really narrow you could probably fit few more columns there 🤔

Jokinen:
You are right that just after `m` there's a screen width that's relatively awkward. The problem is that there's no clear way of resolving the issue without possibly impacting the design. Currently the layout uses the design for desktop and Tapahtumat for mobile views (as there were no mobile views within the design). Because Tapahtumat is a production service, I assumed that this detail is one that's not necessarily worth the time it would take to negotiate a fix.

PR #26, opened 2021-06-09
Liikunta 79 | Add social media share links, map box and other similar section to details page

Jokinen:
Screenshot 2021-06-09 at 12 35 55


tuovinensanttu (approved):
Solid work 👍

PR #27, opened 2021-06-09
Liikunta 75 | Add upcoming events section

Jokinen:
I was able to use columns for this layout after all. The trick was to use `display: inline-block` on the element that appear as a child of the column container.

I also tried to use `flex`, but it would have required a fixed height to be set, which would have made reusability very difficult to do.

I implemented a new card type in order to comply with the desing.

Screenshot 2021-06-09 at 14 52 36

### DIFF
--- a/src/components/card/condensedCard.module.scss
+++ b/src/components/card/condensedCard.module.scss
@@ -26,6 +26,5 @@
 
   @include breakpoints.respond-above(m) {
     padding: $spacing-m !important;
-    padding-top: $spacing-m !important;
   }
 }
--- a/src/components/card/condensedCard.module.scss
+++ b/src/components/card/condensedCard.module.scss
@@ -1,20 +1,31 @@
+@use "breakpoints";
 @import "variables";
 
 .keywords {
-  padding: $spacing-m !important;
-  padding-bottom: 0 !important;
+  @include breakpoints.respond-above(m) {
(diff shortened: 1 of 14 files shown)


src/components/card/condensedCard.module.scss
```diff
@@ -0,0 +1,31 @@
...
+    margin-top: 0.25rem;
+  }
+}
+
+.content {
+  padding-top: $spacing-s !important;
+
+  @include breakpoints.respond-above(m) {
+    padding: $spacing-m !important;
+    padding-top: $spacing-m !important;
```
tuovinensanttu, line 29: `padding-top` is redundant.

tuovinensanttu (commented):
The recommended section is not working quite right on review environment. This happens right after breakpoint (m)
with all the browsers (firefox, chrome, safari)

image

Jokinen:
It took some time to find out the cause. The layout was working with the development build.

I was ultimately able to track down the cause: the declaration for the `columns` rule was incorrect. This caused the contents of the rule to be drawn incorrectly. I had declared count before column width when the correct order is the other way around. I was able to reproduce the error with a local production build, and the fix seemed to resolve the issue.

tuovinensanttu (approved):
This is looking good now 👍  I would have wanted to test this again in review environment, but there is no ETA when it's up and running again.

PR #28, opened 2021-06-16
Liikunta 80 | Integrate data into entity view

Jokinen:
This PR (unfortunately) does a few things

### Adjustment to logic in the venue resolver

I received inconsistent results with some fields. Some missing fields were `null` whereas seemed to be missing entirely. Because our data is spread out between services, it was difficult to find out the root cause. However, I noticed one inconsistency in our end. The `tprek` data transformer provides `null` as the default value (in case the value is missing). With linked data we are using `pick` in which case the data will be `undefined` if it does not exist. By providing default to linked data, the intention is at least clear--that empty values should be represented with `null`. I assumed that this was the intention due to how `tprek` calls had this default value provided for.

### Support for requesting used translations only

I made the venue resolver aware of language. In case a language is provided, the resolver will only return the translations that are requested.

If I did not add this feature, I would have needed to create logic in the UI. If we send translations to the UI, we end up sending data that will never be used, and hence create unnecessary network traffic.

A language can be requested by providing the `accept-language` header.

### Use GraphQL instead of REST

I migrated our REST implementation into GraphQL. Because we use GraphQL with out other integrations, it made sense to not create an exception here.

### Integrate data into entity view

Finally, I integrated the entity view with a data source. Fields which are not available are hidden, other fields should be visible and working.

The "next events" and "other similar sports" sections still use placeholder data.


.github/workflows/review.yml
```diff
@@ -20,6 +20,7 @@ env:
   K8S_READINESS_PATH: /api/readiness
   NEXT_PUBLIC_CMS_GRAPHQL_ENDPOINT: https://liikunta.hkih.production.geniem.io/graphql
   NEXT_PUBLIC_UNIFIED_SEARCH_GRAPHQL_ENDPOINT: https://unified-search.test.kuva.hel.ninja/search
+  NEXT_PUBLIC_NEXT_API_GRAPHQL_ENDPOINT: https://${{ env.K8S_NAMESPACE }}.${{ env.BASE_DOMAIN }}/api/graphql
```
tuovinensanttu, line 23: Review action crashes at the moment with error `The workflow is not valid. .github/workflows/review.yml (Line: 23, Col: 42): Unrecognized named-value: 'env'. Located at position 1 within expression: env.K8S_NAMESPACE`

Seems that this part of the file doesn't have access to `env` and throws an error. Maybe just move the value directly on the lines 37 & 54.

tuovinensanttu (approved):
Looking really nice!

.github/workflows/review.yml
Jokinen: I think I am referencing `env` while defining it, so it makes sense that this approach fails. I couldn't find any DRY approach to this either so inlining sounds like a good option for now.

PR #30, opened 2021-06-23
Liikunta 84 | Show details from hauki on the venue details page

Jokinen:
Adds integration to hauki for the venue query and renders opening hours found from hauki in the venue details view.

Currently hauki integration is only enabled for data that's sourced from `tprek`.

Some linked data may have opening hours. For instance, `linked:25` references the `tprek:25` resource and could be requested by using the id `25`. `linked` prefixed ids do not work with Hauki. However, I am not sure if we can generally expect for opening hours to be found if we use a plain ID parsed from a linked ID. Hence, for now, I've only enabled the integration for `tprek`.

You can get a gist of the intended behaviour when it comes to rendering opening hours by looking at the unit tests for `humanizeOpeningHoursForWeek`. It's noteworthy that the `name` and `description` fields are not rendered in the UI layer. I have omitted them because they are not visible in the designs and I am not sure what content they may contain.

I have also used deep comparisons when grouping times. Perhaps the `periods` field could have been used for this instead. However, I was not able to find documentation for the purpose of this field. It was typed as a string in the API documentation, but seems to be a list of integers in the real world--perhaps a list of IDs that link a time to a period. The deep comparison should be sufficient, but may not be as performant.

Also contains unrelated fixes:
- Fixes the three column layout in venue details view
- Fixes passing of `Accept-Langauge` header
- Improves logging
- Allows developers to see more logging by enabling debug mode

Remaining issues:
1) I was not able to understand how ISR works with languages. It seems that NextJS won't regenerate the venue page on language change. In other words, if the Finnish version is regenerated first, NextJS will server it for the English and Swedish version as well. Any ideas?


tuovinensanttu (approved):
Solid work!

PR #31, opened 2021-06-30
LIIKUNTA-23 | Map view

tuovinensanttu:
Implements map that shows venue locations. By default it shows 10 locations (more if "search more" was pressed on list view before changing views). 

There was also a problem with `react-leaflet` package so I used a fork someone created for the time being. We should change this back to original after this [PR](https://github.com/PaulLeCam/react-leaflet/pull/885) is merged or some other fix is implemented


src/components/list/searchList.module.scss
```diff
@@ -39,7 +39,18 @@
   justify-self: center;
 }
 
+.row {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+}
+
+.showOnMap {
```
Jokinen, line 48: This rule is probably not really used?

tuovinensanttu: Nope, not really sure where I was going with this. I'll remove this.

src/components/search/searchHeader/searchHeader.module.scss
```diff
@@ -0,0 +1,100 @@
...
+      }
+    }
+
+    button {
+      width: 50%;
+
+      @include breakpoints.respond-above(m) {
+        width: 200px;
+        border: 2px solid;
+        height: 45px;
```
Jokinen, line 66: Why `200px` and `45px`?

In essence we are customising HDS so it'd be good to highlight  why so that we don't accidentally discard the customisations.

If you have set a fixed width because you are trying to undo the `width: 50%;`, you could for instance do `width: unset;`.

In this view you are transforming a supplementary HDS button into a secondary HDS button with custom styles. One thing to note is that with supplementary buttons, the button label has no left margin whereas with secondary buttons they do.

Screenshot 2021-07-08 at 11 03 10
Screenshot 2021-07-08 at 11 03 05

src/components/search/searchPageSearchForm/SearchPageSearchForm.tsx
```diff
@@ -1,18 +1,21 @@
...
 import getURLSearchParamsFromAsPath from "../../../util/getURLSearchParamsFromAsPath";
 import Text from "../../text/Text";
 import SuggestionInput, {
   Suggestion,
 } from "../../suggestionInput/SuggestionInput";
 import styles from "./searchPageSearchForm.module.scss";
+import searchApolloClient from "../../../client/searchApolloClient";
+import { getUnifiedSearchLanguage } from "../../../client/utils";
+import { ShowMode } from "../searchHeader/SearchHeader";
+import getShowMode from "../../../util/getShowMode";
```
Jokinen, line 18: To my eye it looks like these imports have been reordered in a weird way.

src/components/search/searchPageSearchForm/SearchPageSearchForm.tsx
```diff
@@ -71,10 +80,19 @@ function SearchPageSearchForm() {
...
-      <form role="search" className={styles.form} onSubmit={handleSubmit}>
+    <div>
+      {showMode === ShowMode.LIST && <Text variant="h1">Mitä etsit?</Text>}
+      <form
+        role="search"
+        className={classNames(styles.form, {
+          [styles.mapMode]: showMode === ShowMode.MAP,
+        })}
+        onSubmit={handleSubmit}
+      >
```
Jokinen, line 95: I think that this component doesn't need to know whether it is being displayed in a list or in a map. It guides future changes in a direction which may make the component harder to maintain because we would be creating a more opaque separation of concerns. I think that this component's job should be to allow the user to input search terms and then refresh the page with the new criteria after the user communicates that they are done.

If there are problems with margin, then the definition of those margins should perhaps be moved to the containing component. If some text needs to be hidden conditionally then we could consider whether we can do that by using a prop.

src/pages/search/index.tsx
```diff
@@ -3,17 +3,22 @@ import { GetStaticPropsContext } from "next";
...
 import Section from "../../components/section/Section";
 import SearchResultCard from "../../components/card/searchResultCard";
 import SearchList from "../../components/list/SearchList";
 import styles from "./search.module.scss";
+import initializeCmsApollo from "../../client/cmsApolloClient";
+import SearchHeader, {
+  ShowMode,
+} from "../../components/search/searchHeader/SearchHeader";
+import getURLSearchParamsFromAsPath from "../../util/getURLSearchParamsFromAsPath";
+import getShowMode from "../../util/getShowMode";
```
Jokinen, line 21: Imports are reordered in an odd way to my eye.

src/pages/search/index.tsx
```diff
@@ -114,24 +133,54 @@ export default function Search() {
...
-          onLoadMore={onLoadMore}
-          count={count}
-          blockSize={BLOCK_SIZE}
-          hasNext={pageInfo?.hasNextPage}
-          items={searchResultItems.map((item) => (
-            <SearchResultCard key={item.id} {...item} />
-          ))}
-        />
-      </Section>
+      <SearchHeader
```
Jokinen, line 161: This is rather specific, but by wrapping the form in a component, it becomes more difficult to reach. This is sort of a pet peeve for me as I've dealt with a few applications with this structure and I find it makes the code harder to grasp to someone who is unfamiliar with it. On this page the integration between the search query and search filter control is rather implicit.

I would prefer that we plug the `SearchForm` into the `SearchHeader` for instance with a prop or through some other means which allows us to reach it from within this component.

src/components/search/searchPageSearchForm/SearchPageSearchForm.tsx
```diff
@@ -71,10 +80,19 @@ function SearchPageSearchForm() {
...
     data?.unifiedSearchCompletionSuggestions?.suggestions ?? [];
 
+  const show = getURLSearchParamsFromAsPath(router.asPath).get("show");
+  const showMode = getShowMode(show);
+
   return (
-    <div className={styles.searchArea}>
-      <Text variant="h1">Mitä etsit?</Text>
-      <form role="search" className={styles.form} onSubmit={handleSubmit}>
+    <div>
```
Jokinen, line 87: Does this div have a purpose?

src/components/search/searchPageSearchForm/SearchPageSearchForm.tsx
```diff
@@ -40,11 +43,17 @@ function SearchPageSearchForm() {
 
   const doSearch = (q?: string) => {
     const nextQuery = q ? { q } : null;
+    const params = getURLSearchParamsFromAsPath(router.asPath);
 
     queryPersister.persistQuery(nextQuery);
-    router.push({ pathname: router.pathname, query: nextQuery }, undefined, {
-      shallow: true,
-    });
+    params.set("q", nextQuery.q);
```
Jokinen, line 49: Here you are accidentally introducing a bug by not persisting all parameters. If the user were to select a place from the map and click back, they would be taken into the list view instead of the map view.

src/pages/search/index.tsx
```diff
@@ -114,24 +133,54 @@ export default function Search() {
...
+    const params = getURLSearchParamsFromAsPath(router.asPath);
+    params.set("show", nextMode);
+
+    router.replace(
+      { pathname: router.pathname, query: params.toString() },
+      undefined,
+      {
+        shallow: true,
+      }
+    );
```
Jokinen, line 154: Personally I would have set up the map view in a different route, e.g. `search/map` for better separation of concerns. Although this approach is nice now in that data can be efficiently shared, the map will likely get its own query in the future, right?

Jokinen (commented):
By and large very high quality set of changes. A few comments owning to the size of the feature.

On mobile the map toggle button and result count don't look finished:
Screenshot 2021-07-08 at 11 39 37

The border between the button in map mode is not a line, because the button's other borders are transparent. You could consider using a pseudo element here:
Screenshot 2021-07-08 at 11 39 58

When in search list mode the background for the search form container is too light:
Screenshot 2021-07-08 at 11 41 05

There's extra padding between the search form container and the results which was introduced in this PR:
Screenshot 2021-07-08 at 11 41 51

You mentioned a few things which should be done separately:
- Accessibility
- Change dependency back into "official" one

Have these been ticketed?

src/components/mapView/mapView.module.scss
```diff
@@ -0,0 +1,17 @@
...
+  width: 100%;
+  z-index: 1;
+
+  @include breakpoints.respond-above(s) {
+    height: calc(100vh - 130px);
+  }
+
+  @include breakpoints.respond-above(m) {
+    height: calc(100vh - 140px);
+  }
```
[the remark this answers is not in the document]
tuovinensanttu: They tweak the MapView in a way that the the map container doesn't overflow the screen. Mainly because the map contains links to openstreet etc, so those are visible all the time.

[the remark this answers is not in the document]
Jokinen: Would it be useful to document this in the file? It would also be interesting to understand where the values come from--why they need to be different for different breakpoints etc.

src/components/search/searchHeader/searchHeader.module.scss
tuovinensanttu: I was indeed trying to undo the `width: 50%`, unset seems to do the trick. Height is set to `45px` because according to design the buttons should be "slimmer". I didn't even notice that the secondary button had left margin, this needs a little tweaking.

src/components/mapView/mapView.module.scss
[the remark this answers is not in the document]
tuovinensanttu: I'll add some comments to the file, but for example breakpoint-s makes the main header slimmer as the logo height gets smaller. Without this the footers container would be shown a little above the buttons and the black border wouldn't cover the whole height.

src/components/search/searchPageSearchForm/SearchPageSearchForm.tsx
tuovinensanttu: There was a "little" merge conflict in the file. While resolving it the import order got messed up.

src/pages/search/index.tsx
tuovinensanttu: Same thing here, these got messed up while I was resolving merge conflicts. I'll fix these as well.

src/components/search/searchPageSearchForm/SearchPageSearchForm.tsx
tuovinensanttu: Yup, without it the search form gets buried under koros while in list view.

src/components/search/searchPageSearchForm/SearchPageSearchForm.tsx
tuovinensanttu: I had a feeling I would mess up something with the query persister 😄 fixing this as well.

src/pages/search/index.tsx
tuovinensanttu: Actually moving map to it's own route would make more sense. Even that we were talking about changing the search for map view it totally skipped my mind while implementing this.

src/pages/search/index.tsx
tuovinensanttu: `SearchForm` is now passed down as a prop.

package.json
```diff
@@ -14,7 +14,10 @@
   },
   "dependencies": {
     "@apollo/client": "^3.3.16",
+    "@monsonjeremy/react-leaflet": "^3.2.2",
+    "@react-leaflet/core": ">=1.0.0 <1.1.0 || ^1.1.1",
```
Jokinen, line 18: Unsually complex version declaration 🤔 I wonder why we can't just use 1.1.1 for instance?

src/components/list/searchList.module.scss
Jokinen: This has not been removed. Was it used after all?

src/pages/search/map.tsx
```diff
@@ -0,0 +1,139 @@
...
+  );
+}
+
+export async function getStaticProps(context: GetStaticPropsContext) {
+  const cmsClient = initializeCmsApollo();
+
+  // We still need to initialize the application with data from the CMS. The
+  // data includes things like menus and languages. If we don't initialize, this
+  // data will be missing when the application is accessed through the search
+  // view.
```
Jokinen, line 128: This comment is perhaps meant for another context.

Jokinen (approved):
Code changes looked good to me 👍

Some visual quirks still remain. E.g. the colour of the search box in list view should be darker and there's an extra margin between the search form and the map toggle button that was not there before (and isn't in the design).

Screenshot 2021-08-16 at 11 39 13
Screenshot 2021-08-16 at 11 41 52

package.json
tuovinensanttu: I should have removed this one, when I ran into the problem that react-leaflet was not working I tried few different fixes that people suggested. Forgot simplify this one.

src/pages/search/map.tsx
tuovinensanttu: Ill remove it.

tuovinensanttu:
I somehow had a feeling that I fixed the colour + margin mishap already. Need to recheck that one.

Jokinen:
I used the PR environment to test the changes. Perhaps it's out of date?

tuovinensanttu:
Hmm, might be. I double check these to make sure.

`showOnMap` is now used. This class makes the "Show on map" button full width on mobile.

tuovinensanttu:
Ahh of course! The new version is not being build because there are conflicts in few files. But the colour + margin should be fixed once conflicts are resolved.

I added simple a link to the popups on map that will navigate user to the venue details page. Those might not be according to design at the moment (I remember seeing a design for them but I can't find it from abstract)

Jokinen:
Looks good now 👍

PR #32, opened 2021-06-30
LIIKUNTA-86 | Link from search results into result page

Jokinen:
1. Adds links to search results so that they link to the venues detail page
2. Replaces placeholder image with image from search data if available
3. Replaces button in SearchResultCard with a fake button to make it easier to use with screen readers
4. Implements the back button in the venue details view--user will be redirected back to search with the search filter they have specified earlier
5. Replace "is router ready" pattern with "is fallback pattern" when waiting for static generation to generate a page
6. Avoid searching on every keystroke. Instead fire search when the user submits the search form.
7. Allow users to reset search by removing the search text


tuovinensanttu (approved):
LGTM 👍

Jokinen:
Rebase to pull CMS integration fixes from main.

PR #33, opened 2021-07-05
Fix CMS integration

Jokinen:
The Graphql CMS API was updated and it broke our integration.

The `dekstopImage` query for landing page does not return the image I expected. I've sent a question about this to Geniem.


tuovinensanttu (approved):
Nice!  👍

PR #34, opened 2021-07-05
Fetch collections from root page

Jokinen:
Don't fetch collections from the connection resolver, instead fetch collections that have been selected to be shown on the front page.

### DIFF
--- a/src/pages/index.tsx
+++ b/src/pages/index.tsx
@@ -2,9 +2,9 @@ import { GetStaticPropsContext } from "next";
 import { NextRouter, useRouter } from "next/router";
 import { gql, useQuery } from "@apollo/client";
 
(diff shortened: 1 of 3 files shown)


tuovinensanttu (approved):
Code wise this is looking good, in review environment collections are rendered as black, I guess thats more a WP problem 🤔

Jokinen:
It's a WP problem in that WP is missing some translations. Can you confirm that they are visible in Finnish, but missing in English?

Of course we could try and hide collections when they are missing all translations. Though I'm not sure it would be worth it. Maybe better to fix the data.

tuovinensanttu:
Yeah, visible in Finnish 👍 I seem to always forgot that the language defaults to english on my browser.

PR #35, opened 2021-07-07
Liikunta-85 | Find ontologies for tprek venues

Jokinen:
In this PR I implement a data enhancer for venue resolving when tprek is used as the source.

Enhancers are data resolvers which can be used to enhance requested data based on the contents of that data. In other words, enhancers depend on data which we have to request from an integration.

In this PR I also refactor the way in which we integrate to other services within the venue query. I attempted to use a more declarative approach in order to bring better clarity about responsibilities and the data we can find based on each integration.

In this PR I also slightly adjust the way in which the Hauki integration works. For some reason the Hauki service is very brittle as of now and almost every other request will fail in a 500 code.

I also adjusted logging to make debugging easier for humans.


tuovinensanttu (approved):
These changes were included in #36 which I wen't through, check comments in that PR.

Jokinen:
I'll resolve your comments on that PR and merge all these changes through that PR.

PR #37, opened 2021-07-09
Fix language changing on venue page

Jokinen:
Previously language changing would not work in the venue page.

After this change data should update when language is changed.


tuovinensanttu (approved):
Changing language works on detail page.

PR #38, opened 2021-07-09
LIIKUNTA-92 | Limit search results to sports locations

Jokinen:
Instead of searching from all locations, search from location which are marked to be under liikunta's "jurisdiction".

In this PR we take into use the `ontologyTreeId` filter and use it to target liikunta locations.

The three looks something like this:

```
KuVa
 - One
 - Two
 - Liikunta (id: 551)
```


Jokinen:
Resolved merge conflict.

PR #40, opened 2021-07-14
LIIKUNTA-18 | Add administrative division filter to search page

Jokinen:
Once fetching for administrative divisions work, this change can be tested and merged.

### DIFF
--- a/src/components/search/searchPageSearchForm/searchPageSearchForm.module.scss
+++ b/src/components/search/searchPageSearchForm/searchPageSearchForm.module.scss
@@ -8,6 +8,15 @@
   grid-gap: $spacing-m;
 
   @include breakpoints.respond-above(s) {
-    grid-template-columns: 3fr 1fr;
+    grid-template-columns: repeat(4, 1fr);
+
+    & > *:first-child,
+    & > *:nth-child(2) {
+      grid-column: span 2;
+    }
+  }
+
+  .submitButton {
+    grid-column: 4;
   }
 }
--- a/src/components/search/searchPageSearchForm/searchPageSearchForm.module.scss
+++ b/src/components/search/searchPageSearchForm/searchPageSearchForm.module.scss
@@ -16,7 +16,9 @@
     }
   }
 
-  .submitButton {
-    grid-column: 4;
+  @include breakpoints.respond-above(s) {
+    .submitButton {
+      grid-column: 4;
+    }
   }
 }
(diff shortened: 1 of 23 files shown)


Jokinen:
Resolved merge conflicts and improved i18n support.

src/components/search/searchPageSearchForm/searchPageSearchForm.module.scss
```diff
@@ -8,6 +8,15 @@
...
+    grid-template-columns: repeat(4, 1fr);
+
+    & > *:first-child,
+    & > *:nth-child(2) {
+      grid-column: span 2;
+    }
+  }
+
+  .submitButton {
+    grid-column: 4;
```
tuovinensanttu, line 20: This should be inside the breakpoint definition above. As it is now, it will break the mobile layout


image

src/widgets/administrativeDivisionDropdown/AdministrativeDivisionDropdown.tsx
```diff
@@ -0,0 +1,86 @@
...
+      defaultValue={defaultValue}
+      label={label}
+      clearButtonAriaLabel="Poista kaikki valinnat"
+      selectedItemRemoveButtonAriaLabel="Poista ${value}"
+      options={options}
+      onChange={handleOnChange}
+      placeholder={placeholder}
+      icon={<IconLocation />}
+    />
+  );
```
tuovinensanttu, line 85: There is open discussion in [slack](https://helsinkicity.slack.com/archives/C01DWA87NJC/p1629444430026600), that the component should be similar to one tapahtumat.hel.fi uses. Based on that this component should be `Combobox` instead of `Select`. Ofc we can disable the multiselect option for the time being, as US doesn't support it yet.

Jokinen:
Fixed the mobile layout issue @tuovinensanttu mentioned and changed the input into a combobox.

I also sorted and filtered the returned divisions based on the discussion on Slack. Perhaps this behaviour will move into the search backend at some point at least partially, but I thought that the fastest way to a working implementation was to use a temporary change in the UI layer.

Jokinen:
Integrated the division control into the map view as well by sharing filtering logic between the list and map views. Also modified styles slightly by removing `overflow-y: hidden` so that the division dropdown can overlap the search header.

A few venues looked to be missing locations so I added logic into the map view which avoids rendering an item if it does not have a location.

tuovinensanttu (approved):
Looks good to me 👍

Jokinen:
Rebased to resolve conflicts.

PR #41, opened 2021-07-15
LIIKUNTA-96 | Add collections page

Jokinen:
In this PR I add the collections page and implement a way to render the HTML that the CMS sends.

For the collection page I've not yet implemented event listing. It was unclear how these events would be sourced or where they should link to. Perhaps this would require a new query to be added--in which case it's better to do separatrely.


tuovinensanttu (approved):
Looked this through last week, forgot to add a review. IMO these changes are looking good 👍

Jokinen:
Resolved merge conflict.

PR #43, opened 2021-08-18
Use a grid instead of columns for upcoming events

Jokinen:
Use a grid instead of columns to get a left-to-right order instead of a column order for upcoming events.


tuovinensanttu (approved):
Upcoming events are now nicely ordered by date.

PR #44, opened 2021-08-20
LIIKUNTA-111 | Support event searches on collection page

Jokinen:
Adds support for the "event search" module on the collection page.

Also implements pagination for contents of collections.

I wanted to use relay styled pagination because it's still the recommendation although its status is a bit unclear. It was made difficult by the fact that linked events uses a page number based pagination system. In hindsight perhaps it would have been better to not opt to use relay style here.


Jokinen:
Force pushed to resolve merge conflicts to keep on top of them.

Jokinen:
Fixed issue with the pagination that caused the last page to have non-working a11y support on the collections page.

public/locales/sv/collection_page.json
```diff
@@ -1,3 +1,7 @@
 {
-  "share_sport": "SV: Jaa liikunta"
+  "share_sport": "SV: Jaa liikunta",
+  "show_more_events": "SV: Näytä lisää",
+  "no_more_events": "SV: Ei enempää tapahtumia",
+  "loading_more_events_label": "EV: Ladataan lisää",
```
tuomas777, line 5: Engdish? 😉

tuomas777 (approved):
Verified that pagination a11y works now, and all in all this is very great 💯 🚀

PR #46, opened 2021-08-23
LIIKUNTA-69 | Add no results message

Jokinen:
Adds an empty search notice when there are no results for a search.

Screenshot 2021-08-23 at 13 02 03
Screenshot 2021-08-23 at 13 01 27

### DIFF
--- a/src/components/list/SearchList.tsx
+++ b/src/components/list/SearchList.tsx
@@ -1,9 +1,10 @@
 import React, { Ref, forwardRef } from "react";
 import classNames from "classnames";
-import { Button, LoadingSpinner, IconMap } from "hds-react";
+import { Button, LoadingSpinner, IconMap, IconSearch } from "hds-react";
 
 import styles from "./searchList.module.scss";
 import Text from "../text/Text";
+import InfoTemplate from "../infoTemplate/InfoTemplate";
 
 type Props = {
   items: React.ReactElement[];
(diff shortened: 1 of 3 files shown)


src/components/list/SearchList.tsx
```diff
@@ -32,6 +33,16 @@ const SearchList = forwardRef(
     const a11yIndex = (Math.floor(items.length / blockSize) - 1) * blockSize;
     const loadedMoreAmount = hasNext ? blockSize : count - items.length;
 
+    if (count === 0) {
+      return (
+        <InfoTemplate
+          icon={IconSearch}
+          title="Sorry no search results were found for the search criteria you selected"
+          description="Try changing your search criteria and you're sure to find something to do!"
+        />
```
tuovinensanttu, line 42: Minor nitpicky comment: All the other placeholder texts in the application are in finnish, so this being only one in english causes it to stick out.

tuovinensanttu (approved):

src/components/list/SearchList.tsx
Jokinen: The design had an English translation so that's why--I didn't want to go about and translate it myself as there was talk that we should avoid our own translations and leave that work to translators or content producers.

I agree with you though. My thinking was that as I was going to implement translations next, this inconsistency would not be in the application for long.

PR #47, opened 2021-08-24
LIIKUNTA-108 | Implement translations

Jokinen:
Some content was not translated ("micro copy").

This PR adds the `next-i18next` and uses it to provide translations for static non data sourced content.

Translations are sourced from the content management system. They are fetched into the project with `get-translations`.

`next-i18next` handles bundling of necessary translations with each page request. Required translations should be communicated on the page level by using `serverSideTranslationsWithCommon` or `serverSideTranslations`.

Currently translations are copied into version control. It's possible to source them directly from the CMS over the network by configuring a different backend for `i18next`.


tuomas777 (approved):
Looks very good, works as it should, and could not spot even a single missing translated string 🙂 Great job!! 🥇

PR #48, opened 2021-08-25
LIIKUNTA-124 | Fix search term not resetting in the UI

Jokinen:
Previously the search term would not reset on language change. This change forces the value to synchronise with search parameter values. The approach is not ideal, but I would not think of another small fix to resolve the issue.


tuomas777 (approved):
It seems the bug isn't reproducible at the moment for some reason, but of course feel free to merge this anyway if you think this will be more robust 🙂

PR #49, opened 2021-08-25
LIIKUNTA-126 | Fix map searches redirecting to list view

Jokinen:
Previously searches in the map view of search would redirect the user into the search view. After this change map view searches will stick to the map view.


tuomas777 (approved):
Works as expected 👍

PR #50, opened 2021-08-25
LIIKUNTA-127 | Fix apollo client being created multiple times

Jokinen:
Fixes issue where a reference was naively mutated from a parameter value. In JS the reference the original value holds can not be changed in this way.

This resulted in the created client not being cached, which meant that it would be created on every request for it. This is the intended behaviour on server side, but on client side the client can be reused for the duration of the session.

Not sure what the practical implications were exactly. I would assume that as the client in essence resets, the cache within does as well, which results in some duplicate data between views being available later (e.g. menus, languages, etc).

In this PR I add a `MutableReference` class to fix the issue and to make this part of the code less oblique and more robust.

FYI @Sartonon , thanks for spotting this!


tuomas777 (approved):
Looks good AFAICT 👍

PR #51, opened 2021-08-25
LIIKUNTA-107 | Fix scroll position when returning to search

Jokinen:
The user can return to the search list by two means
* Using the back button of their browser
* Using the back button in the application UI

**Browser back button**

If the user uses their browser, we shouldn't manage scrolling. Instead the browser should do it for us. There may be some caveats with this behaviour, but broadly speaking a custom implementation would not be better compared to letting the browsers do their best.

To fix scrolling behaviour with the back button, I've included the pagination information (`after`, `first`) in the search parameters. This way we can store and rediscover them by using the same mechanism we use with other search parameters. This should allow Apollo to correctly find the previous query from cache and almost instantly open the previous page--which allows the browser scrolling behaviour to take place correctly.

There's one caveat and that is that the back button does not work in `dev` mode. In `dev` mode each page is forcefully rebuilt every time it's accessed. This causes the loading of the page to stall, which means that the browser scrolls the view to the correct position before the result list has been rendered. This will break scrolling behaviour.

**UI back button**

When the user makes use of the back button in the UI, we can't rely on the browser to do the scrolling for us. Instead we have to implement our own logic for scrolling the view.

I've used an approach where the VenuePage provides a `scrollTo` direction to the SearchPage by including it in the search params it sends to the SearchPage.  SearchPage completes a `querySelector` query over the list items it renders with the value provided in `scrollTo`. The querying happens outside of React in the DOM. If an element is found, its `scrollIntoView` method is used to bring it into view. If it's not found, nothing happens.

The `scrollTo` parameter is lost after the user completes a new query or loads more results by using the pagination.


tuovinensanttu (commented):
Tested this locally and with review environment.
- Locally functionality seems to be working as expected.
- In review environment functionality works with first "load more" action. After that returning from detail page shows 20 results and scroll ends in a wrong position.

I also noticed that the `first` and `after` url params never change, shouldn't these be updated as well when user loads more results (at least the `after` param)?

I tried to add a console log to `search/index` on line 119 to check the `after` value, it seems that the value never changes so maybe something iffy going on with US?

Jokinen:
Yes, the `after` parameter should change. The `first` parameter in essence means page size and it doesn't need to change necessarily.

The differences in the values that come from US are very slight:

eyJvZmZzZXQiOjIwfQ==
eyJvZmZzZXQiOjMwfQ==

It's still possible there's something off with them.

I was able to reproduce the behaviour and will try to find out the cause.

Jokinen:
Rebased to resolve conflicts, because I remembered that they'd be more invasive.

I think the issue was that the new pagination variables were overwritten by the old ones due to an incorrect spread order. Both current search params and the new pagination params are stored in the query parameters. Because search params were applied last, the `after` param within them overwrote the new `after` param.

In a previous version, `useUnifiedSearchParameters` did not return pagination variables.

tuovinensanttu (approved):
Seems to be working now in review. Nice job! 👍

PR #54, opened 2021-08-30
LIIKUNTA-112 | Add event count labels to collection cards on home page

Jokinen:
There are two kinds of event modules
* selected
* search

The selected module contains a list of IDs that refer to events. Getting the count for these is easy--just count the amount of IDs.

The search module is more complex. It only contains information about the query parameters for finding a group of events. To find out the total count, we first have to request more information from linked (through our own BFF API).

While the event count is being determined, there should be a loading indicator.

The procedure for finding out the event count should fire after the event grid enters the view. This should defer the firing of the requests for the event counts so that queries which are needed for rendering before the fold content can take precedence.

Screenshot 2021-08-30 at 12 45 54


tuovinensanttu (approved):
Looking good, only minor thing I noticed is that on the event page when pressing load more, page is scrolled to first list item.

Jokinen:
You mean the collection details page event lists?

I'm currently trying to resolve that issue in LIIKUNTA-118. For some reason it's proving difficult to fix.

PR #55, opened 2021-08-31
LIIKUNTA-99 | Make map view more accessible.

tuovinensanttu:
This PR makes map view is more accessible (or little less as keyboard navigation is disabled 😅 little confused here). Added a wrapper for MapContainer that should tell user that they have focused a map and it's meant for visual users only.


Jokinen:
Looks good to me 👍

Jokinen (approved):

PR #56, opened 2021-09-02
LIIKUNTA-119 | Hide footer on map view

tuovinensanttu:
Added a prop to `Page` component that hides the footer. Passing a prop through one extra component might not be ideal but imo it's clearer than adding a direct check to `Layout` component that checks if the url path matches the map view.


Jokinen (approved):
Looks good 👍

PR #57, opened 2021-09-03
LIKUNTA-115 | Add accessibility statement and about us pages

tuovinensanttu:
This PR adds accessibility statement and about us pages. I'm still wondering if I should have created one component that eats `PAGE_QUERY` as the data is almost identical between the pages 🤔.

### DIFF
--- a/src/components/footer/Footer.tsx
+++ b/src/components/footer/Footer.tsx
@@ -72,10 +72,11 @@ function Footer({ navigationItems }: Props) {
         copyrightHolder={t("copyright_holder")}
         copyrightText={t("copyright_text")}
       >
-        <HDSFooter.Item href="/about" label={t("about_us")} />
+        <HDSFooter.Item href="/about" label={t("about_us")} as={Link} />
         <HDSFooter.Item
           href="/accessibility"
           label={t("accessibility_statement")}
+          as={Link}
         />
       </HDSFooter.Base>
     </HDSFooter>
--- a/src/components/footer/Footer.tsx
+++ b/src/components/footer/Footer.tsx
@@ -72,8 +72,11 @@ function Footer({ navigationItems }: Props) {
         copyrightHolder={t("copyright_holder")}
         copyrightText={t("copyright_text")}
       >
-        <HDSFooter.Item href="#" label={t("about_us")} />
-        <HDSFooter.Item href="#" label={t("accessibility_statement")} />
+        <HDSFooter.Item href="/about" label={t("about_us")} />
+        <HDSFooter.Item
+          href="/accessibility"
+          label={t("accessibility_statement")}
+        />
       </HDSFooter.Base>
     </HDSFooter>
   );
--- a/src/pages/about/index.tsx
+++ b/src/pages/about/index.tsx
@@ -1,0 +1,78 @@
+import { useRouter } from "next/router";
+import { gql, useQuery } from "@apollo/client";
+import { GetStaticPropsContext } from "next";
+
+import HtmlToReact from "../../components/htmlToReact/HtmlToReact";
+import initializeCmsApollo from "../../client/cmsApolloClient";
+import getPageMetaPropsFromSEO from "../../components/page/getPageMetaPropsFromSEO";
+import { getQlLanguage } from "../../client/utils";
+import { LANDING_PAGE_QUERY } from "../index";
+import seoFragment from "../../domain/seo/cmsSeoFragment";
+import Page from "../../components/page/Page";
+import serverSideTranslationsWithCommon from "../../domain/i18n/serverSideTranslationsWithCommon";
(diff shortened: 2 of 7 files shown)


src/pages/accessibility/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+    variables: {
+      languageCode: language,
+    },
+  });
+
+  return {
+    props: {
+      initialApolloState: cmsClient.cache.extract(),
+      ...(await serverSideTranslationsWithCommon(context.locale, [
+        "home_page",
```
Jokinen, line 73: Unused translations.

src/components/footer/Footer.tsx
```diff
@@ -72,8 +72,11 @@ function Footer({ navigationItems }: Props) {
         copyrightHolder={t("copyright_holder")}
         copyrightText={t("copyright_text")}
       >
-        <HDSFooter.Item href="#" label={t("about_us")} />
-        <HDSFooter.Item href="#" label={t("accessibility_statement")} />
+        <HDSFooter.Item href="/about" label={t("about_us")} />
+        <HDSFooter.Item
+          href="/accessibility"
+          label={t("accessibility_statement")}
+        />
```
Jokinen, line 79: Using `as={LInk}` would translate the paths they point to (which you configured in the `i18nRoutes.config.js` file).

Jokinen (approved):
The components could be combined. The slug for them could be set as dynamic and provided as a variable

```
  query PageQuery($languageCode: LanguageCodeEnum!, $uri: !String) {
    page(id: $uri, idType: URI) {
```

It would introduce some extra considerations: how to ensure that there are no URL collisions, and if there are, how do we resolve them. For instance, perhaps we should use a `pages` directory for these kind of pages to give them their own namespace.

But it would allow us to define this integration once and would guide us to maintain the integration in a way that is in line with the thinking of the headless WP API (as far as I am aware).

Personally I would consider combining the pages if time allows--better to have less code to maintain. But I am also fine with merging this PR as it is, and combining the pages later when we know more.

Jokinen:
One thing I am thinking is whether we should have moved this particular menu into the CMS at this point: https://liikunta.content.api.hel.fi/wp-admin/nav-menus.php?action=locations

Perhaps we could have used a tertiary menu for this? Not entirely sure about the purpose for it. I assume that primary is the main menu and secondary is the menu in the footer.

src/pages/about/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  );
+}
+
+export async function getStaticProps(context: GetStaticPropsContext) {
+  const cmsClient = initializeCmsApollo();
+  const language = getQlLanguage(context.locale ?? context.defaultLocale);
+
+  await cmsClient.pageQuery({
+    nextContext: context,
+    query: LANDING_PAGE_QUERY,
```
Jokinen, line 63: Incorrect query--breaks caching.

src/pages/accessibility/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  );
+}
+
+export async function getStaticProps(context: GetStaticPropsContext) {
+  const cmsClient = initializeCmsApollo();
+  const language = getQlLanguage(context.locale ?? context.defaultLocale);
+
+  await cmsClient.pageQuery({
+    nextContext: context,
+    query: LANDING_PAGE_QUERY,
```
Jokinen, line 63: Incorrect query--breaks caching.

Jokinen (changes requested):

src/pages/about/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+    variables: {
+      languageCode: language,
+    },
+  });
+
+  return {
+    props: {
+      initialApolloState: cmsClient.cache.extract(),
+      ...(await serverSideTranslationsWithCommon(context.locale, [
+        "home_page",
```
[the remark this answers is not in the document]
tuovinensanttu: Hmm, I'm not sure if I did something wrong. But without this the translations in the main navigation are missing / using translation key.

src/pages/about/index.tsx
```diff
@@ -60,7 +59,7 @@ export async function getStaticProps(context: GetStaticPropsContext) {
 
   await cmsClient.pageQuery({
     nextContext: context,
-    query: LANDING_PAGE_QUERY,
+    query: ACCESSIBILITY_PAGE_QUERY,
```
Jokinen, line 62: Is this query named in an odd way? Considering that this is the about page?

Jokinen (approved):

src/pages/about/index.tsx
tuovinensanttu: Renamed now. Seems that I can't win with these copy pasted things on this PR 😄

PR #59, opened 2021-09-07
LIIKUNTA-118 | Fix wrong scroll position when more events are loaded

Jokinen:
The load more announcer element was positioned unexpectedly which caused unexpected scroll behaviour.

Because the announcer element was absolutely positioned within the grid, its position on the page was outside the event list itself:
Screenshot 2021-09-07 at 8 56 35
(Highlighted line is position of the announcer element)

I used the `preventScroll` option of the focus method to prevent scrolling on focus. Evergreen browsers should have support for it.


tuovinensanttu (approved):
LGTM 👌

PR #60, opened 2021-09-07
LIIKUNTA-138 | Add clustering to map view

Jokinen:
POI that are close to each other on the map are now clustered. By clicking on the cluster, the map is zoomed. Clusters are also unclustered when the user zooms the map.

Also stops using fork of `react-leaflet` as the application now supports webpack 5.

Screenshot 2021-09-07 at 11 22 47
Screenshot 2021-09-07 at 11 23 15

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,6 @@
   },
   "dependencies": {
     "@apollo/client": "^3.3.16",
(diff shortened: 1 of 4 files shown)


tuovinensanttu (approved):
Works great! Nico job!

PR #62, opened 2021-09-08
LIIKUNTA-142 | Add filter list into search view

Jokinen:
Adds a filter list into the search view.

All selected filters should be shown on a list that allows for removing them.

Array based filters will be unwrapped so that each value in the array get its own item in the filter list.

Contains refactoring to the unifiedSearch module.

Screenshot 2021-09-08 at 15 15 50
Screenshot 2021-09-08 at 15 17 28


tuovinensanttu (approved):
Really clean PR, especially the `useUnifiedSearch` hook wraps all the functionality nicely. 🏅

tuovinensanttu:
Actually there seems to be a bug when I tried to remove administrative division filter. Keyword label is removed + URL is changed correctly, but when I press "Search" again, Keyword + url parameter gets applied back.

PR #63, opened 2021-09-09
LIIKUNTA-139 | Add accessibility sentences

tuovinensanttu:
This PR implements accessibility sentences to API and to Venue page.


src/components/infoBlock/InfoBlock.tsx
```diff
@@ -93,6 +93,42 @@ function InfoBlockList({ items, inline }: InfoBlockContentListProps) {
...
 
+type InfoBlockCollapseProps = {
+  icon: React.ReactElement;
+  items: React.ReactElement[];
+  className?: string;
+  title: string;
+  titleClassName?: string;
+};
+
+function InfoBlockCollapse({
```
Jokinen, line 104: HDS now has the Accordion component, and with it they introduced the `useAccordion` hook for custom accordion components. Maybe we should use that so that we have better guarantees that the accordion is accessible?

https://hds.hel.fi/components/accordion#custom-accordion

src/domain/graphql/venue/venueSchema.ts
```diff
@@ -62,6 +67,7 @@ const typeDefs = gql`
     isOpen: Boolean
     ontologyTree: [Ontology]!
     ontologyWords: [Ontology]!
+    accessibilitySentences: [AccessibilitySentences]
```
Jokinen, line 70: I think that the array can be required here (`!`) as you are using an empty array as a default value in `src/domain/graphql/venue/venueResolver.ts`.

src/components/mapBox/MapBox.tsx
```diff
@@ -1,17 +1,20 @@
...
 import styles from "./mapBox.module.scss";
+import { AccessibilitySentences } from "../../types";
 
 type Props = {
   title: string;
   serviceMapUrl: string;
   placeName: string;
   placeAddress: string;
   links?: React.ReactElement<React.ComponentProps<typeof InfoBlock.Link>>[];
+  accessibilitySentences?: AccessibilitySentences[];
```
Jokinen, line 17: I think that passing the accessibility data as is does not necessarily make sense here as you've gone through the effort of implementing the `Collapse` content model for `InfoBlock`. Instead of passing `accessibilitySententences`, perhaps it should be an `InfoBlock.Collapse` element that gets passed.

I'd also consider whether, depending on what your intention was, we should only have a single prop of type `InfoBlockContent[]`, which then allows us to pass the links and the collapsed content.

src/types.ts
```diff
@@ -164,6 +175,7 @@ export type VenueDetails<T = TranslationsObject> = {
   telephone: T | null;
   ontologyTree: Ontology[];
   ontologyWords: Ontology[];
+  accessibilitySentences: AccessibilityTranslationsObject;
```
Jokinen, line 178: Is this type definition correct? The generics use with this type allows for it to be used to represent a value that contains all the translations, as well as a value where a translation has been picked (`VenueDetails<T = TranslationsObject>`). Will the type of `accessibilitySentences` always be of type `AccessibilityTranslationsObject`?

Jokinen (approved):
The way in which the multiline content is indented does not quite match the design

Design:
Screenshot 2021-09-13 at 9 58 11

Code:
Screenshot 2021-09-13 at 9 57 14

src/components/infoBlock/InfoBlock.tsx
tuovinensanttu: That makes sense. I totally forgot that HDS had that component.

src/types.ts
tuovinensanttu: Ahh, seems that I misunderstood the type definition. In that case the accessibility sentences should be `AccessibilityTranslationsObject | AccessibilitySentences`

PR #64, opened 2021-09-13
LIIKUNTA-133 | Allow multiple admin divisions to be targeted

Jokinen:
Changes the admin division select into a multiselect.

Changes the filtering logic for the combobox suggestions. Now allows partial matches. E.g. "töölö" will match "Etu-Töölö".


tuovinensanttu (commented):
Removing values from combobox (clear all / remove single from `Keyword`) doesn't update filters (black keywords) even if search button is pressed again.

image

Jokinen:
@tuovinensanttu I changed the logic around filter modifying and the bug should now be removed.

tuovinensanttu (approved):

PR #67, opened 2021-09-20
LIIKUNTA-125 | Add keyword parameter to events query.

tuovinensanttu:
Adds keyword parameter to `linked events` events query. 

Testing:
1) [Review](https://liikunta-helsinki-67.test.kuva.hel.ninja/paikat/tprek:239)
2) [Staging](https://liikunta-helsinki.test.kuva.hel.ninja/paikat/tprek:239)

Seems that the keyword is filtering at least some events, in the testing case "Karanneet eläimet" event should not be shown in the Review environment.

### DIFF
--- a/src/domain/graphql/event/eventSchema.ts
+++ b/src/domain/graphql/event/eventSchema.ts
@@ -10,7 +10,7 @@ const typeDefs = gql`
     language: String
     text: String
     translation: String
-    keywords: [String]
+    keywords: [String!]
   }
 
   extend type Query {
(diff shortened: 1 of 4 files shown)


src/domain/graphql/event/eventSchema.ts
```diff
@@ -10,6 +10,7 @@ const typeDefs = gql`
     language: String
     text: String
     translation: String
+    keywords: [String]
```
Jokinen, line 13: Maybe do `String!` to disallow falsy values in the keywords list.

Jokinen (approved):
Looks good 👍

PR #69, opened 2021-09-27
LIIKUNTA-82 | As a citizen I want to see locations that are open now

Jokinen:
Adds an "open now" filter, which limits results to places that have opening hours information available and are currently open according to it.


tuovinensanttu (approved):
LGTM

PR #71, opened 2021-09-30
LIIKUNTA-156 & LIIKUNTA-167 | Show all results in map view

tuomas777:


src/pages/search/map.tsx
```diff
@@ -44,14 +43,26 @@ export const MAP_SEARCH_QUERY = gql`
...
+              sv
+              en
+            }
+            location {
+              geoLocation {
+                geometry {
+                  coordinates
+                }
+              }
+            }
```
Jokinen, line 60: I understand the purpose here and I think it's worth it because we are asking for so much data in one go. This does mean that we can't reuse the fetched data in other views, and that other views will have to request the more complete set of data in a new request. It's not worth it to preload additional fields from thousands of results in order to avoid making the additional query though.

Jokinen (approved):
Codewise this looks good and the intended outcome on the map works well too.

I think that the intention was that the link from this PR is provided to Pekka, Päivi and Tomi before merging. But insofar as I can say, this looks good to me!

PR #72, opened 2021-10-02
Fix scrolling issue when loading more events

Sartonon:
## Description :sparkles:

When clicking load more button in the search page, it causes page to scroll to top and then right back to load more announcer. Also same request is done twice to API when fetching more event with load more button, I think it has something to do with `router.replace` and useQuerys fetchPolicy settings. To fix this issue, I added possibility to add `TransitionOptions` to setFilters function. Also unifiedSearchQuery wasn't using variables for caching because of `relayStylePagination`. I think we can get better cahcing by using keyArgs and getting rid of non-default fetching policies.

- Added possibility to pass TransitionOptions to setFilters function
- Deleted fetch policies from useUnifiedSearchQuery apollo hook 
- Added `excludeArgs` utility to be used with `relayStylePagination`

## Testing :alembic:

Click "Näytä lisää sijainteja" button on search page and page should be scrolled to top and then back down.

### DIFF
--- a/src/common/apollo/utils.ts
+++ b/src/common/apollo/utils.ts
@@ -2,6 +2,16 @@ import { ApolloClient, NormalizedCacheObject } from "@apollo/client";
 
 import { Connection, MenuItem } from "../../types";
 
+export const excludeArgs =
+  (excludedArgs: string[]) =>
+  (
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    args: Record<string, any> | null
+  ) =>
+    args
+      ? Object.keys(args).filter((key: string) => !excludedArgs.includes(key))
+      : false;
+
 export function getNodes<T>(connection: Connection<T>): T[] {
   return connection.edges.map(({ node }) => node);
 }
--- a/src/common/apollo/utils.ts
+++ b/src/common/apollo/utils.ts
@@ -3,11 +3,7 @@ import { ApolloClient, NormalizedCacheObject } from "@apollo/client";
 import { Connection, MenuItem } from "../../types";
 
 export const excludeArgs =
-  (excludedArgs: string[]) =>
-  (
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    args: Record<string, any> | null
-  ) =>
+  (excludedArgs: string[]) => (args: Record<string, unknown> | null) =>
     args
       ? Object.keys(args).filter((key: string) => !excludedArgs.includes(key))
       : false;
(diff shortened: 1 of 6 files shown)


src/domain/unifiedSearch/useUnifiedSearchQuery.ts
```diff
@@ -64,8 +64,6 @@ export default function useUnifiedSearchQuery<TData = any>(
       ...searchParams,
       ...variables,
     },
-    fetchPolicy: "cache-and-network",
-    nextFetchPolicy: "cache-only",
```
Sartonon, line 68: @Jokinen  How does this exactly work? I think this had issues with our setFilters implementation.

Sartonon: Ok found out that the searh is actually broken without these. No idea why...

Sartonon: I figured that the problem is with the `relayStylePagination` helper which doesn't use any keyArgs for chaching by default. So I added `excludeArgs(["after"])` to it and now it seems to work nicely without:

```
fetchPolicy: "cache-and-network",
    nextFetchPolicy: "cache-only",
```

Jokinen: I think that `nextFetchPolicy: "cache-only"` was implemented to allow for nicer behaviour when using the back arrow or browser's back button to navigate back from a venue page into the search result list. It's meant to return the previous content as quickly as possible when the user navigates back, so that the page contains all the elements the user previously viewed, and the browser is able to scroll the user to the previous position in the list.

The `fetchPolicy` is likely earlier code and may address a different issue.

Sartonon: I think that with this PR it should be still working like that

src/common/apollo/utils.ts
```diff
@@ -2,6 +2,16 @@ import { ApolloClient, NormalizedCacheObject } from "@apollo/client";
 
 import { Connection, MenuItem } from "../../types";
 
+export const excludeArgs =
+  (excludedArgs: string[]) =>
+  (
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    args: Record<string, any> | null
```
Jokinen, line 9: I assume unknown didn't pass type checking here?

Jokinen (approved):
Code looked good, and tested the broken behaviour--it seems to be working now. Good job!

src/common/apollo/utils.ts
Sartonon: Hmm need to check. This was actually just copy paste of what I had done in the other project

Sartonon: changed any to unknown

PR #75, opened 2021-10-05
LIIKUNTA-181 | Fix tags in results page and add ontologyWordIds query

Sartonon:
## Description :sparkles:

- Fix search result tags when clicking them
- Add ontologyWordsIds query

### Closes :no_good_woman:

**[LIIKUNTA-181](https://helsinkisolutionoffice.atlassian.net/browse/LIIKUNTA-181):**


src/common/components/card/Card.tsx
```diff
@@ -223,8 +227,13 @@ function Card({ children, id, className }: CardProps) {
 
   const handleWrapperMouseUp = (e: React.MouseEvent<HTMLElement>) => {
     const link = linkRef.current;
+    const keywords = keywordsRef.current;
     const up = new Date();
 
+    // ignore if click event comes from keywords (allow default behaviour)
+    if (keywords !== e.target && keywords.contains(e.target as Node)) {
+      return;
+    }
```
Jokinen, line 236: This is a good way to resolve the problem here, but I feel like it reveals some downsides in the original implementation. The logic for deciding what clicks "count" entangles the child components with this parent one. Not all cards have keywords and perhaps some will have other clicks targets in the future. Adding refs for all doesn't create clear separation of concerns in my opinion and breaks composability to some degree.

Could we for instance do `even.stopPropagation()` in the mouseUp handler of the element the `keywordsRef` references? Or could we pass some custom field in the event which would mark it as to be ignored in this handler? Don't really have a clear suggestion in mind here to be honest, there seem to be downsides with every approach 🤔

Sartonon: I see where you are coming from. And yes this feels little bit clunky but I also don't see huge problem here. Of course it is one option to add mouseUp handler to keywords container like you said.

But honestly I dont see a problem with the "entanglement". I think that the components are already entangled and are meant to be used together. And in this case Card acts as a parent component that handles some of the logic :)

Sartonon: Updated the implementation:

```
   // stop propagation so link click is not simulated in Card component
  const handleKeywordMouseUp = (e: React.MouseEvent<HTMLLIElement, MouseEvent>) => {
    e.stopPropagation()
  }
   <li key={key} className={styles.keyword} onMouseUp={handleKeywordMouseUp}>
```

Sartonon: not necessarily final

Jokinen: It's not an issue I see as vital to fix, it's more of an unrefined detail. It may be better to wait it out and I'm happy to go with any decision you make.

When it comes to entanglement, I'm talking about the sort of two way binding that implicitly happens that I'm referencing. The card component is relatively difficult to maintain, because it has many variants and contexts where it is being used. Because of that I think it's important to have simple and clear responsibilities between the different components within the module as well. Usually more contained is also simpler, and easier to maintain and compose. But I do agree with you in that it doesn't make as much of a difference because it's sort of an implementation detail and doesn't impose anything on code that uses the card component.

Jokinen (approved):
Code wise this looks good to me.

I'd capitalize the label that's shown in the filter list
Screenshot 2021-10-06 at 9 52 40

nikomakela (approved):

PR #77, opened 2021-10-06
Fix footer koros style

nikomakela:
Footer had extra line between koros and solid bottom area.

LIIKUNTA-182

image

image

### DIFF
--- a/src/common/components/footer/footer.module.scss
+++ b/src/common/components/footer/footer.module.scss
@@ -6,7 +6,8 @@
   grid-column: 1 / -1 !important;
 
   // koros stuff
-  margin-top: -1rem;
+  margin-top: -14px;
+
   > div > svg {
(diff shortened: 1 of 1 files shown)


src/common/components/footer/footer.module.scss
```diff
@@ -6,7 +6,8 @@
   grid-column: 1 / -1 !important;
 
   // koros stuff
-  margin-top: -1rem;
+  margin-top: -14px;
```
Jokinen, line 9: How were you able to determine this value? is it something we could document here?

nikomakela: It's the height of koros

nikomakela: -0.9em works the same, but since the koros was defined with pixels, I used them.

PR #79, opened 2021-10-07
Fix/liikunta 194 collection focus

Sartonon:


src/domain/pagination/usePagination.tsx
```diff
@@ -23,42 +14,31 @@ type A11yHelpers = {
...
-        // a grid.
-        moreResultsAnnouncerRef.current.focus({ preventScroll: true });
-      }
-
-      return result;
-    },
-    [fetchMoreBase, moreResultsAnnouncerRef]
-  );
+}: Options): A11yHelpers {
+  useLayoutEffect(() => {
```
Jokinen, line 23: I think that using an `effect` here is suspect as we have a clear user action that we can use as a source for this change (user clicking the load more button). Can you open your thinking behind this decision? I would have expected that we can get correct DOM values as we are in an event handler.

Sartonon: I need to do little bit thinking on this but I thought layoutEffect needs to be used because then the list has been rerendered (not painted yet) and we can resolve the correct position of the announcer.

PR #82, opened 2021-10-12
LIIKUNTA-154 | Add new map page for venue and link to it from venue page

Sartonon:
Added new page for showing single venue in a map. Ended up placing the new venue map page to `/pages/venues/[id]/map/index.tsx`. Map can be accessed from venue page.

## Screenshot
Screenshot 2021-10-12 at 18 11 49

### DIFF
--- a/src/common/components/mapView/mapView.module.scss
+++ b/src/common/components/mapView/mapView.module.scss
@@ -8,6 +8,7 @@
   height: calc(100vh - 120px);
   width: 100%;
   z-index: 1;
+  position: relative;
 
   // At breakpoint-s the logo shrinks a little, so we need to adjust map size to match that
   @include breakpoints.respond-above(s) {


src/common/components/infoBlock/InfoBlock.tsx
```diff
@@ -73,7 +73,7 @@ function InfoBlockLink({
   }
 
   return (
-    <Link href={href}>
+    <Link href={href} avoidEscaping>
```
Sartonon, line 76: This one was little bit of a pain :) Needed to add avoidEspacing for the Link to work properly.

src/pages/venues/[id]/map/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+export default function VenueMapPage(props) {
+  const nextApiApolloClient = useNextApiApolloClient(
+    props.initialNextApiApolloState
+  );
+
+  return (
+    <Page
+      title="Liikunta-Helsinki"
+      description="Liikunta-helsinki"
```
Comment (Sartonon) [inline src/pages/venues/[id]/map/index.tsx:86]: Seems like venue page is also missing proper title and description. @Jokinen Do you have good ideas what to do with these?

src/pages/venues/[id]/map/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+  try {
+    await cmsClient.pageQuery({
+      nextContext: context,
+    });
+    await nextApiClient.query({
+      query: VENUE_QUERY,
+      variables: {
+        id: context.params.id,
+      },
+    });
```
Comment (Sartonon) [inline src/pages/venues/[id]/map/index.tsx:116]: Deleted headers part from here:

```
   context: {
        headers: {
          "Accept-Language": context.locale ?? context.defaultLocale,
        },
      },
```

Is it needed to avoid problems somewhere?

src/common/components/infoBlock/InfoBlock.tsx
Jokinen: This should be maybe applied where you use the component, e.g. in `venues/[id]/index.tsx`. Or if you prefer it as a default, then consider toggling it on by default in the `Link` component.

src/common/components/page/Page.tsx
```diff
@@ -36,7 +36,7 @@ export const PAGE_FRAGMENT = gql`
 
 export const PAGE_QUERY = gql`
   ${PAGE_FRAGMENT}
-  query PageQuery($language: LanguageCodeFilterEnum) {
+  query PageQuery($menuLocation: MenuLocationEnum!) {
```
Jokinen, line 39: 👍

src/pages/venues/[id]/map/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+import styles from "./map.module.scss";
+
+const MapView = dynamic(
+  () => import("../../../../common/components/mapView/MapView"),
+  {
+    ssr: false,
+  }
+);
+
+const VenueMapPageContent: React.FC = () => {
```
Comment (Jokinen) [inline src/pages/venues/[id]/map/index.tsx:27]: You are probably aware, but using `React.FC` is a topic of a bit of division. Our team stopped using it a year ago when CRA did: https://github.com/facebook/create-react-app/pull/8177

src/pages/venues/[id]/map/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+    },
+  };
+
+  return (
+    <div className={styles.content}>
+      <div className={styles.mapHeader}>
+        <Link
+          href={backHref}
+          avoidEscaping
+          aria-label="Takaisin paikkanäkymään"
```
Comment (Jokinen) [inline src/pages/venues/[id]/map/index.tsx:61]: Missing translation.

You may also need to apply this label on the anchor directly. NextJS may not propagate this prop.

src/pages/venues/[id]/map/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+        >
+          <a>
+            <IconArrowLeft aria-hidden="true" size="l" />
+          </a>
+        </Link>
+        <h1 className={styles.title}>{title}</h1>
+      </div>
+      <MapView
+        zoom={20}
+        center={[location[1], location[0]]}
```
Comment (Jokinen) [inline src/pages/venues/[id]/map/index.tsx:71]: Minor comment, but you could destructure these earlier on in which case you could name them and be more explicit about which one is lat and which one is long. For whatever reason, they often end up getting mixed.

src/pages/venues/[id]/map/map.module.scss
```diff
@@ -0,0 +1,28 @@
+@use "breakpoints";
+@import "variables";
+
+.content {
+  --background-color: #{$color-engel-medium-light};
+  grid-column: 1 / -1 !important;
+}
+
+.title {
+  font-size: 24px;
```
Comment (Jokinen) [inline src/pages/venues/[id]/map/map.module.scss:10]: Could use a design token from HDS.

Applies more generally to this file as well, `1rem` `$spacing-m` and `color: black` could be `$color-black`.

src/pages/venues/[id]/map/map.module.scss
```diff
@@ -0,0 +1,28 @@
...
+  margin-left: var(--spacing-xl);
```
Comment (Jokinen) [inline src/pages/venues/[id]/map/map.module.scss:11]: We've used SASS variables because it allows us to lock HDS design tokens into a specific version. Because CSS variables are global, they can be altered by dependencies. Downside is that you lose themability interop.

Not a major point of concern for me, FYI.

src/pages/venues/[id]/map/map.module.scss
```diff
@@ -0,0 +1,28 @@
...
+}
+
+.mapHeader {
+  display: flex;
+  align-items: center;
+  background-color: var(--background-color);
+  padding: 0 1rem;
+  height: 70px;
```
Comment (Jokinen) [inline src/pages/venues/[id]/map/map.module.scss:19]: I was wondering whether we should declare this as `rem` to give a better sense what the value is in design system units?

src/pages/venues/[id]/map/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+const VenueMapPageContent: React.FC = () => {
+  const router = useRouter();
+  const { data, loading, error } = useQuery(VENUE_QUERY, {
+    variables: {
+      id: router.query.id,
+    },
+  });
+
+  if (loading || error) {
```
Comment (Jokinen) [inline src/pages/venues/[id]/map/index.tsx:35]: I'm wondering about this 🤔 In essence it shouldn't really matter in this case because (1) the data should be available when we first render the map in the client and (b) refocusing the map could be awkward UX. But generally speaking, I think, that we should render the map as early as possible so that the images etc. have as much time as possible to load.

src/pages/venues/[id]/map/index.tsx
Comment (Sartonon) [reply src/pages/venues/[id]/map/index.tsx:61]: Yeah, I need to learn to use WP translations next :)

src/pages/venues/[id]/map/index.tsx
Comment (Sartonon) [reply src/pages/venues/[id]/map/index.tsx:71]: I will do that

src/common/components/infoBlock/InfoBlock.tsx
Jokinen: I didn't want to break the default behaviour of `Link`. I doubt that it'll break anything, but there may be niche cases where the behaviour differs from that of Next.

When you toggle this prop on, it takes the `object` representation of a route and stringifies it this side instead of within the library. In the library the params get encoded, which causes issues. We don't do that in the custom logic.

That may lead to inconsistent behaviour in some cases.

PR #85, opened 2021-10-14
LIIKUNTA-208 | Add ability to focus single venue in map search

Sartonon:
Added abilito to click "Näytä kartalla" link in venue card in search page. It opens map search page with selected venue focused and zoomed in, also marker popup is opened by default.

In summary

- Use `venue` query parameter to focus single venue in map search page
- Restore scroll position when changing back to list page

### DIFF
--- a/src/common/components/mapView/MapView.tsx
+++ b/src/common/components/mapView/MapView.tsx
@@ -99,6 +99,7 @@ function MapView({ items = [], center, zoom, focusedItemId }: Props) {
           iconCreateFunction={createCustomClusterIcon}
           maxClusterRadius={60}
           animate={false}
+          key={items.join() ?? "default"}
         >
           {items.map((item) => {
             if (!item.location) {
--- a/src/pages/search/map.tsx
+++ b/src/pages/search/map.tsx
@@ -159,15 +159,12 @@ export default function MapSearch() {
           <SearchPageSearchForm showTitle={false} searchRoute="/search/map" />
         }
       />
-      {showVenueFocusedMap ? (
-        <MapView
-          key="focusedMap"
-          items={searchResultItems}
-          focusedItemId={venueId}
-        />
-      ) : (
-        <MapView items={searchResultItems} focusedItemId={venueId} />
-      )}
+      <MapView
+        // Use key to force rerender when mode is changes. Otherwise map position / zoom wouldn't reset
+        key={showVenueFocusedMap ? "focused-venue-map" : "search-map"}
+        items={searchResultItems}
+        focusedItemId={showVenueFocusedMap ? venueId : undefined}
+      />
     </Page>
   );
 }


src/common/components/mapView/MapView.tsx
```diff
@@ -45,10 +46,32 @@ type Props = {
...
   const { t } = useTranslation("map_view");
+  const focusedItem = useMemo(
+    () => items.find((item) => item.id === focusedItemId),
+    [items, focusedItemId]
+  );
+
+  const getFocusedItemPosition = (): LatLngExpression | null => {
+    if (focusedItem && focusedItem.location) {
+      const lat = focusedItem.location[1];
+      const lng = focusedItem.location[0];
```
Jokinen, line 62: Suggestion

```suggestion
      const [lng, lat] = focusedItem.location;
```

src/common/components/mapView/MapView.tsx
```diff
@@ -78,16 +101,28 @@ function MapView({ items = [], center, zoom }: Props) {
...
+            // used to automatically open popup for focused map item
+            // react-leaflet doesn't seems to provide better wway to do this
+            const handleOpenPopup = (marker) => {
+              const isFocusedItem = focusedItemId && item.id === focusedItemId;
+              if (isFocusedItem) {
+                // doesn't work without setTimeout. There seems to be some asynchronity
+                // in leaflet libary when rendering markers.
+                setTimeout(() => marker?.openPopup());
+              }
+            };
```
Jokinen, line 118: This is likely something the React API doesn't provide for and you would need to integrate into Leaflet directly. Maybe it'd be possible to do with `Map.load` for instance. You can find the map instance with `useMap`. Just context, I don't think you need to change the current approach, it works for me.

src/common/components/mapView/MapView.tsx
```diff
@@ -78,16 +101,28 @@ function MapView({ items = [], center, zoom }: Props) {
...
             if (!item.location) {
               return null;
             }
 
+            // used to automatically open popup for focused map item
+            // react-leaflet doesn't seems to provide better wway to do this
+            const handleOpenPopup = (marker) => {
+              const isFocusedItem = focusedItemId && item.id === focusedItemId;
+              if (isFocusedItem) {
+                // doesn't work without setTimeout. There seems to be some asynchronity
```
Jokinen, line 114: Missing capital letter.

src/pages/search/map.tsx
```diff
@@ -146,7 +159,15 @@ export default function MapSearch() {
...
-      <MapView items={searchResultItems} />
+      {showVenueFocusedMap ? (
+        <MapView
+          key="focusedMap"
+          items={searchResultItems}
+          focusedItemId={venueId}
+        />
+      ) : (
+        <MapView items={searchResultItems} focusedItemId={venueId} />
+      )}
```
Jokinen, line 170: What's happening here? To the naked eye it looks like the conditional isn't needed here.

src/common/components/mapView/MapView.tsx
```diff
@@ -45,10 +46,32 @@ type Props = {
...
-function MapView({ items = [], center, zoom }: Props) {
+function MapView({ items = [], center, zoom, focusedItemId }: Props) {
   const { t } = useTranslation("map_view");
+  const focusedItem = useMemo(
+    () => items.find((item) => item.id === focusedItemId),
+    [items, focusedItemId]
+  );
+
+  const getFocusedItemPosition = (): LatLngExpression | null => {
+    if (focusedItem && focusedItem.location) {
```
Jokinen, line 60:
```suggestion
    if (focusedItem?.location) {
```

src/common/components/mapView/MapView.tsx
```diff
@@ -45,10 +46,32 @@ type Props = {
   items: MapItem[];
   center?: LatLngExpression;
   zoom?: number;
+  focusedItemId?: string;
 };
 
-function MapView({ items = [], center, zoom }: Props) {
+function MapView({ items = [], center, zoom, focusedItemId }: Props) {
```
Jokinen, line 52: With this API change the `MapView` component entangles more business logic. It's something I'm okay with, but personally I would've attempted to keep that away from this component insofar as it's possible. E.g. provide the zoom as a prop and open the marker in a handler.

It's a little bit of extra work and I'm not entirely sure how it would work out so this approach may be more pragmatic in the end, which is why I think it to be a good choice as well.

Jokinen (approved):
Looked to be working as promised when tested. Small code comments.

src/pages/search/map.tsx
Jokinen: Maybe this has something to do with forcing the map to re-render?

PR #87, opened 2021-10-19
LIIKUNTA-210 | Fix broken filter keywords

Jokinen:
The issues were caused by behaviour that differs from the default behaviour of NextJS and were brought about with the change where we begun to use our custom stringification routine by default. Unfortunately it didn't have support for `query` as a way of providing search parameters, nor did it retain the current `pathname` if a new one was not provided. The stringification routine Next uses is relatively complex and this regression highlights the difficulty of maintaining logic that attempts to mimic that. We should look to replace the current `escape` with a better approach, but I didn't spend time on that now.

Screenshot 2021-10-19 at 16 28 08

### DIFF
--- a/src/domain/i18n/router/Link.tsx
+++ b/src/domain/i18n/router/Link.tsx
@@ -7,31 +7,14 @@ import useRouter from "./useRouter";
 import { getI18nPath, stringifyUrlObject } from "./utils";
 import { Locale } from "../../../config";
 
-function getI18nHref(
-  href: string | UrlObject,
-  locale: Locale,
-  defaultPathname: string
-): string | UrlObject | null {
-  if (typeof href === "string") {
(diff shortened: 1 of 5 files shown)


src/domain/i18n/router/Link.tsx
```diff
@@ -7,30 +7,14 @@ import useRouter from "./useRouter";
...
 
-function getHrefThatAvoidsEscaping(
-  href: string | UrlObject | null
-): string | null {
-  if (!href || typeof href === "string") {
-    // If the href is a string it won't be unescaped
-    return null;
-  }
-
+function getHrefThatAvoidsEscaping(href: UrlObject | null) {
```
Sartonon, line 17: Should we still keep this function or just use `stringifyUrlObject`?

Sartonon (approved):
Great :)

PR #89, opened 2021-10-20
LIIKUNTA-207 | Fix date time picker closing on helper body click

Jokinen:
## Description

After this change, clicks within the dropdown of DateTimePIcker should not close the picker.

Other changes:
- I removed an unused import that was left in by accident earlier
- I renamed the test file and very slightly adjusted the existing test case for DateTimePicker

### Closes

**[LIIKUNTA-207](https://helsinkisolutionoffice.atlassian.net/browse/LIIKUNTA-207):**

### Automated tests

I attempted to write automated tests for this case, because it's easy for a dev to not notice a regression, but I couldn't get event propagation to work in a way that would have allowed mimicing the actual usecase.

### Manual testing

I have tested the change manually.

### DIFF
--- a/src/common/components/dateTimePicker/DateTimePicker.tsx
+++ b/src/common/components/dateTimePicker/DateTimePicker.tsx
@@ -212,6 +212,8 @@ export default function DateTimePicker({
   const handleWrapperOnBlur = (e: FocusEvent) => {
     if (
       isOpen &&
+      // Progress only is a relatedTarget exists. Otherwise we can't check
+      // whether it exists within currentTarget.
       e.relatedTarget &&
       !e.currentTarget.contains(e.relatedTarget as Node)
     ) {


src/common/components/dateTimePicker/DateTimePicker.tsx
```diff
@@ -213,6 +213,7 @@ export default function DateTimePicker({
     if (
       process.browser &&
```
Sartonon, line 214: Started wondering if this is needed 🤔

src/common/components/dateTimePicker/DateTimePicker.tsx
```diff
@@ -213,6 +213,7 @@ export default function DateTimePicker({
...
       isOpen &&
+      e.relatedTarget &&
```
Sartonon, line 216: Should this have some comment explaining why this is needed? Developer might think that this is redundant because it is not easy to see the difference in behaviour of the dropdown. Especially because we don't have automated tests for this.

PR #90, opened 2021-10-29
LIIKUNTA-192 | Update packages

Sartonon:
## Description

Update project packages. Most notably update next to version 12 and start using the Rust compiler instead of babel.

I was able to update all packages expect eslint and react-leaflet. Eslint 8 is not yet supported by Next and react-leaflet had strange error:

Screenshot 2021-10-29 at 11 36 51

### Closes

**[LIIKUNTA-192](https://helsinkisolutionoffice.atlassian.net/browse/LIIKUNTA-192):**


src/domain/pagination/types.ts
```diff
@@ -1,21 +1,15 @@
 import {
   ApolloQueryResult,
-  DocumentNode,
   FetchMoreOptions,
   FetchMoreQueryOptions,
-  TypedDocumentNode,
+  OperationVariables,
 } from "@apollo/client";
 
 // Copied from Apollo's type definitions
```
Jokinen, line 8: Is this copied from Apollo?

src/pages/api/graphql.ts
```diff
@@ -58,7 +58,9 @@ const apolloServer = new ApolloServer({
   dataSources,
   typeDefs,
   resolvers,
-  tracing: process.env.NODE_ENV !== "production",
+  // TODO: do we need this? Apollo server v3 removed this but can be used this way:
+  // plugins: [require('apollo-tracing').plugin()]
+  // tracing: process.env.NODE_ENV !== "production",
```
Jokinen, line 63: It has been used to trace API calls that have a waterfall structure and may take some time. Doesn't need to be enable always as far as I am considered.

src/pages/api/graphql.ts
```diff
@@ -68,11 +70,19 @@ const apolloServer = new ApolloServer({
...
+}
 
 export const config = {
   api: {
     bodyParser: false,
   },
 };
 
-export default apolloServer.createHandler({ path: "/api/graphql" });
+// export default apolloServer.createHandler({ path: "/api/graphql" });
```
Jokinen, line 88: Probably meant to be deleted.

Jokinen (commented):
Changes look good. Answered a few questions.

The review environment doesn't seem to work:
https://liikunta-helsinki-90.test.kuva.hel.ninja/en

I haven't seen an error message like this before. Seems like it doesn't come from next, but it doesn't look like on our infra has given out before.

I wonder if something is off or whether the environment has expired in some funny way? I started the build actions again now.

src/domain/pagination/types.ts
Sartonon: not 1:1 copy but still mostly copied from there

Jokinen (approved):
Looks good to me. Swift 🚀

PR #95, opened 2021-12-09
LIIKUNTA-212 | Fix non visible popup when focusing clustered venue

Jokinen:
## Description

Venues that are clustered on max zoom did not previously open their popup when they were focused.

This change invokes `zoomToShowLayer` before opening the popup, which "unravels" (spiderifies) clustered markers, allowing for the popup invocation to open the popup.

### Closes

**[LIIKUNTA-212](https://helsinkisolutionoffice.atlassian.net/browse/LIIKUNTA-212)**


Sartonon (approved):
Nice 👍

PR #97, opened 2021-12-14
fix broken venue ids in venue map page

Sartonon:
## Description

- Use venue id from query to fix issues with unsupported venue data sources

### Closes

**[LIIKUNTA-213](https://helsinkisolutionoffice.atlassian.net/browse/LIIKUNTA-213):**


Jokinen (approved):
Looks good, good job 👍

Tested with `paikat/tprek:45901`

PR #98, opened 2021-12-15
LIIKUNTA-196 | Order by name when filters are empty

Jokinen:
## Description

Fixes issue where a search with empty filters would change `orderBy` into `relevance`. After this change, only searches with filters should change `orderBy` into `relevance`.

### Closes

**[LIIKUNTA-196](https://helsinkisolutionoffice.atlassian.net/browse/LIIKUNTA-196)**


Sartonon (approved):
Looks good but I noticed that there are some inconsistencies with how `OrderBy.relevance` is added to the query. For example when deleting filters, order by filter changes.

Jokinen:
Thanks for testin thoroughly 👍 I fixed this issue, order should persist over filter removals. When all filters are cleared, it should now reset.

PR #100, opened 2021-12-22
LIIKUNTA-197 | Show opening hours for selected date

Jokinen:
## Description

Modifies the `InfoBlock` for opening hours on the search list to show opening hours for the selected date.

Refactors unified search integration to allow for more accurate typings.

### Closes

**[LIIKUNTA-197](https://helsinkisolutionoffice.atlassian.net/browse/LIIKUNTA-197):**

## Screenshots

Screenshot 2021-12-22 at 17 05 43
Screenshot 2021-12-22 at 17 05 57
Screenshot 2021-12-22 at 17 07 45


src/pages/search/index.tsx
```diff
@@ -139,13 +103,66 @@ function getIsOpenNow(times: Time[]): boolean {
...
+    return null;
+  }
+
+  return (
+    <InfoBlock
+      target="card"
+      headingLevel="h3"
+      icon={<IconClock aria-hidden="true" />}
+      name={
+        !(openAt && openAtFilterIsToday) && isOpenNow
```
Sartonon, line 148: Rather complex condition inlined here. At least I can't easily understand this 🤔

src/pages/search/index.tsx
```diff
@@ -139,13 +103,66 @@ function getIsOpenNow(times: Time[]): boolean {
...
+}: OpeningHoursInfoBlockProps): React.ReactElement<
+  React.ComponentProps<typeof InfoBlock>
+> | null {
+  const { t } = useTranslation("search_page");
+
+  const openingHourTimes =
+    (openAt
+      ? openingHours?.data?.find(
+          (timeData) => timeData.date === format(openAt, "yyyy-MM-dd")
+        )?.times
```
Sartonon, line 125: These brackets confuse me. Aren't they redundant?

Sartonon (commented):
Looks good, generated types are great addition :) Some comments left

src/pages/search/index.tsx
Sartonon: Is it possible for `isOpenNow` to be `true` but we wouldn't render open now label?

Jokinen: I spent some time with this, but I couldn't find a clearer way.

I'll make a helper function so that we can multiline it and maybe make it easier to read that way.

When an `openAt` filter is selected, we want to use that date even if the place would be ope now.

src/pages/search/index.tsx
Jokinen: If I understand correctly, it was unclear where the bracket before `openAt` ends. The bracket exists so that we can use a default value for the entire Elvis expression:

```
(openAt
      ? oneValue
      : twoValue) ?? []
```

Instead of using a single variable, I used two to remove the complex brackets.

Sartonon: Yeah you are correct here, brackets were needed :)

Sartonon (approved):
Great!

PR #103, opened 2022-01-03
FIX LIIKUNTA-158 | Fix location selection links

Jokinen:
## Description

Previously the id from `tprek` was used as is. The id also contained its source (for instance `lipas:...`) which caused the site to build an incorrect link.

After this change the source information is ignored and it's replaced with `tprek`.

Contains a slight refactor for how parts of a venue id are parsed.

### Closes

**[LIIKUNTA-158](https://helsinkisolutionoffice.atlassian.net/browse/LIIKUNTA-158)**

### Manual testing

This collections contains a venue selection: http://localhost:3000/kokoelmat/meno-ja-meininkia-viikonlopuksi

## Screenshots

Screenshot 2022-01-03 at 14 18 00

### DIFF
--- a/src/domain/unifiedSearch/useUnifiedSearchListQuery.ts
+++ b/src/domain/unifiedSearch/useUnifiedSearchListQuery.ts
@@ -1,15 +1,20 @@
 import {
   useSearchListQuery,
   SearchListQueryVariables,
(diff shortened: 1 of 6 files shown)


Sartonon (approved):
Looks good :)

PR #104, opened 2022-01-10
LIIKUNTA-259 | Fix ontologies overlapping datetime picker

Jokinen:
## Description

Changes z-index so that date time picker isn't rendered below keywords.

### Closes

**[LIIKUNTA-259](https://helsinkisolutionoffice.atlassian.net/browse/LIIKUNTA-259):**

## Screenshots

Screenshot 2022-01-10 at 14 04 17
Screenshot 2022-01-10 at 14 04 21

### DIFF
--- a/src/common/components/dateTimePicker/dateTimePicker.module.scss
+++ b/src/common/components/dateTimePicker/dateTimePicker.module.scss
@@ -54,7 +54,7 @@
       padding: 1.2rem 1.2rem;
       left: 0;
       right: 0;
-      z-index: 100;
+      z-index: 1000;
       display: flex;
       flex-flow: column;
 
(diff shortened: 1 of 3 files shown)


Sartonon:
Having issues running this branch locally.

Screenshot 2022-01-12 at 17 20 56

Sartonon:
Something going on with collections:

```
 graphQLErrors: [
    {
      debugMessage: 'Cannot return null for non-nullable field "Collection.id".',
      message: 'Internal server error',
      extensions: [Object],
      locations: [Array],
      path: [Array]
    }
  ],
```

Sartonon (approved):
Looks good but while trying to figure out why this has to be done or why keywords have high z-index I noticed that some event card look strange in venue page:

Screenshot 2022-01-13 at 9 36 53

Jokinen:
They do look strange, but some of that is due to the data being a bit difficult to navigate. The amount of tags is limited to an amount that would look good when the keyword has a one/two word label.

Jokinen:
I added a fix to the collection issue you mentioned.

PR #107, opened 2022-01-19
Move runtime depedency into main deps

Jokinen:
## Description

Fixes issue in staging that causes next server to crash.


pjlampil (approved):
LGTM 👍

PR #109, opened 2022-01-24
Always import useTranslation from next-i18next

Jokinen:
## Description

Always imports `useTranslation` from `next-i18next` so that the hooks has the correct translation context.

### DIFF
--- a/src/common/components/dateTimePicker/DateTimePicker.tsx
+++ b/src/common/components/dateTimePicker/DateTimePicker.tsx
@@ -16,7 +16,7 @@ import {
   IconCross,
 } from "hds-react";
 import classNames from "classnames";
-import { useTranslation } from "react-i18next";
+import { useTranslation } from "next-i18next";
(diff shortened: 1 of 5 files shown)


src/common/components/dateTimePicker/DateTimePicker.tsx
```diff
@@ -141,7 +141,8 @@ export default function DateTimePicker({
   const dateString = getDate(value);
   const timeString = getTime(value);
 
-  const { t } = useTranslation("date_time_picker");
+  const { t, i18n } = useTranslation("date_time_picker");
+  console.log("mooo", i18n);
```
Sartonon, line 145: `mooo`

Sartonon (approved):
Delete `console.log`. Otherwise looks good :D

src/common/components/dateTimePicker/DateTimePicker.tsx
Jokinen: 🐮

PR #111, opened 2022-01-27
LIIKUNTA-269 | Fix language changing problem in venue page

Sartonon:
## Description

Changing language didn't change the language of venue's fields on the venue page. Main issue was that we were using same apollo client in server side between requests. I think new apollo client with empty cache should be created with every revalidation. Now we can still use `Accept-Language` header to request different languages.

### Closes

**[LIIKUNTA-269](https://helsinkisolutionoffice.atlassian.net/browse/LIIKUNTA-269):**


Jokinen (approved):
Nice solution 👍 Good job.

Code looks good and I tested the UI a bit to verify that it works.

PR #112, opened 2022-01-28
LIIKUNTA-270 | Adjust venue grid layout ot match the design

Sartonon:
## Description

Adjusted venue page layout to match design. Not sure if "Muuta samankailtaista liikuntaa" section would also need some adjustment.

### Closes

**[LIIKUNTA-270](https://helsinkisolutionoffice.atlassian.net/browse/LIIKUNTA-270):**

## Screenshots

Screenshot 2022-01-28 at 11 33 41

### DIFF
--- a/src/pages/venues/[id]/venue.module.scss
+++ b/src/pages/venues/[id]/venue.module.scss
@@ -42,8 +42,9 @@
   @include breakpoints.respond-above(m) {
     grid-template-columns:
       [control] 2.5rem
-      [content] 1fr
-      [aside] 1fr;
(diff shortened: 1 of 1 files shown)


Jokinen:
The alignment of the section you mentioned seems to keep with the design. Although I do agree in that it doesn't align that nicely. If it were to be more narrow, we'd most likely have to use 3 instead of 4 items

Screenshot 2022-01-31 at 12 26 49

Jokinen (approved):
Fix works nicely, good job!

PR #115, opened 2022-02-01
LIIKUNTA-222 | Add matomo tracking

Sartonon:
## Description

Created Matomo component that handles user tracking

### Closes

**[LIIKUNTA-222](https://helsinkisolutionoffice.atlassian.net/browse/LIIKUNTA-222):**

### Manual testing

With each page / url change there should be HTTP call to digiariis domain.

Screenshot 2022-02-01 at 9 38 22


src/domain/matomo/__tests__/__snapshots__/matomo.test.jsx.snap
```diff
@@ -0,0 +1,9 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`PageWrapper matches snapshot 1`] = `
+<div>
+  <div>
+    Children
+  </div>
+</div>
```
Jokinen: Should the snapshot look like this? What does this test verify?

src/domain/matomo/matomo.tsx
```diff
@@ -0,0 +1,43 @@
+import {
```
Jokinen: Looks like version control may have this file name in lowercase.

src/domain/matomo/matomo.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const getMatomoUrlPath = (path: string) =>
+  `${process.env.NEXT_PUBLIC_MATOMO_URL_BASE}${path}`;
+
+const matomoInstance = createMatomoInstance({
+  disabled: process.env.NEXT_PUBLIC_MATOMO_ENABLED !== "true",
+  urlBase: process.env.NEXT_PUBLIC_MATOMO_URL_BASE as string,
+  srcUrl: getMatomoUrlPath("piwik.min.js"),
+  trackerUrl: getMatomoUrlPath("tracker.php"),
+  siteId: Number(process.env.NEXT_PUBLIC_MATOMO_SITE_ID),
```
Jokinen: We've accessed environment variables through `Config` so that they are easier to refactor.

src/domain/matomo/matomo.tsx
```diff
@@ -0,0 +1,43 @@
...
+});
+
+const Matomo: React.FC = ({ children }) => {
```
Jokinen: We've used `function Matomo(` styled components in this codebase.

src/pages/_app.tsx
```diff
@@ -7,14 +7,15 @@ import { LoadingSpinner } from "hds-react";
 import Error from "next/error";
 import { appWithTranslation } from "next-i18next";
 import { ToastContainer } from "react-toastify";
-import "nprogress/nprogress.css";
 
+import "nprogress/nprogress.css";
```
Jokinen, line 11: Unexpected that this import has jumped into the in-app exports.

.github/workflows/review.yml
```diff
@@ -38,6 +41,9 @@ jobs:
           DOCKER_BUILD_ARG_NEXT_PUBLIC_CMS_GRAPHQL_ENDPOINT: ${{ env.NEXT_PUBLIC_CMS_GRAPHQL_ENDPOINT }}
           DOCKER_BUILD_ARG_NEXT_PUBLIC_UNIFIED_SEARCH_GRAPHQL_ENDPOINT: ${{ env.NEXT_PUBLIC_UNIFIED_SEARCH_GRAPHQL_ENDPOINT }}
           DOCKER_BUILD_ARG_NEXT_PUBLIC_NEXT_API_GRAPHQL_ENDPOINT: ${{ env.CUSTOM_ENVIRONMENT_URL }}/api/graphql
+          DOCKER_BUILD_ARG_NEXT_PUBLIC_MATOMO_URL_BASE: "//webanalytics.digiaiiris.com/js/"
+          DOCKER_BUILD_ARG_NEXT_PUBLIC_MATOMO_SITE_ID: 587
+          DOCKER_BUILD_ARG_NEXT_PUBLIC_MATOMO_ENABLED: "false"
```
Jokinen, line 46: We've used 0/1 flags for boolean variables because other projects have preferred it previously.

Jokinen (approved):
Slight code style comments. Mainly things that don't align with stylistic choices in this codebase. I pointed them out, you can decide how you want to, if you want to, resolve them.

Otherwise this looks sound, good job!

src/domain/matomo/__tests__/__snapshots__/matomo.test.jsx.snap
Sartonon: Yes. Just a very simple test that makes sure that Matomo component renders the children. Not saying it is very important / ciritcal test but doesn't hurt to have it :)

src/domain/matomo/matomo.tsx
Sartonon: Fixed casing with `git mv`. Seems like it is not changed just by changing it from vscode, exprienced this problem in the past too.

src/pages/_app.tsx
Sartonon: Hmm, interesting that eslint didn't complain about that one, works on other imports

Jokinen (approved):

src/config.ts
```diff
@@ -60,6 +60,27 @@ class Config {
     };
   }
 
+  static get matomoConfiguration() {
+    const matomoUrlBase = this.getEnvOrError(
```
Jokinen, line 64: If this is an optional config, we don't need to error. E.g. if the application can function without it, we do not need to error.

Without some API configs, the application will be broken and useless.

Sartonon: Ill check if it is optional config

Sartonon: Matomo config seems to require URL. Do you think we should add some logic to not include Matomo at all if for example the URL enrivonment variables is missing?

Screenshot 2022-02-01 at 12 18 37

Sartonon: Maybe the simplest solution would be to just have the Matomo enabled if all three env variables are defined. Not sure if it is even necessary to have any of those variables other than  `NEXT_PUBLIC_MATOMO_ENABLED` in environment variables

Jokinen: I'm fine with configuring each environment with the necessary variables and then using matomo's configs to ensure that we don't push stats from other than production environments.

I added the original comment because the CI is breaking on missing env variables. If we should always provide them, then we should configure the CI env to have them.

I mean it'd be ideal if it weren't included when there are no configurations so that we can ensure we don't accidentally log something from an environment that's irrelevant. However, as I recall that's difficult and matomo doesn't provide an easy way to do it out of the box.

Jokinen (approved):

PR #118, opened 2022-02-01
LIIKUNTA-171 | Update documentation for devs and slightly refactor for clarity

Jokinen:
## Description

- Updates readme so it's up to date and inline with best practices

Also contains some other changes that came up while writing the documentation:

- Allows default revalidation time to be configured
- Deprecates `LiikuntaApolloClient` and instead uses a helper function which aligns page rendering between pages
- Removes mock file that wasn't used
- Stores page level tests separately from test utils in order to avoid confusion
- Enables default logging formatter in production
- Refactors application levels configs--stores them under `domain/app` and adds tests

### Closes

**[LIIKUNTA-171](https://helsinkisolutionoffice.atlassian.net/browse/LIIKUNTA-171)**


src/domain/app/getLiikuntaStaticProps.ts
```diff
@@ -0,0 +1,110 @@
...
+  try {
+    await getGlobalCMSData({ client: cmsClient, context });
+    const result = await tryToGetPageProps({ cmsClient });
+    const props =
+      "props" in result
+        ? {
+            initialApolloState: cmsClient.cache.extract(),
+            ...result.props,
+          }
+        : undefined;
```
Sartonon, line 46: Does this mean that if props are not returned from callback, initialApolloState is not extracted?

Sartonon (approved):
Stellar :) Looks very, very good 👍  

`getLiikuntaStaticProps` is nice addition!

src/domain/app/getLiikuntaStaticProps.ts
Jokinen: The return value of a `getStaticProps` function looks like this:

```
export type GetStaticPropsResult<P> =
  | { props: P; revalidate?: number | boolean }
  | { redirect: Redirect; revalidate?: number | boolean }
  | { notFound: true; revalidate?: number | boolean }
```

Which leads me to believe, that for a type correct result, we can't inject any fields into `props` if the field does not exist in the return type, because in that case its a return value is of some type that does not have a props field.

As far as I can tell, if the props field is missing, the result has to be a redirect or a not found variant. I assumed that the props will then be extracted when the user lands on a page that does provide props.

src/domain/app/getLiikuntaStaticProps.ts
```diff
@@ -0,0 +1,110 @@
...
+  };
+  const { data } = await client.query(queryOptions);
+
+  client.writeQuery({
+    ...queryOptions,
+    data: {
+      ...data,
+      // Only use languages that this project supports
+      pageLanguages: getSupportedLanguages(data.pageLanguages, context),
+    },
```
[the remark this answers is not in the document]
Jokinen: Hmn, I think that it's possible.

Spent some time thinking about some way around this, but couldn't come up with one that would just be an adjustment.

A `read` function in the cache was the first thing to come to mind, but I'm not sure if that can be done neatly due to the structure of this data.

Maybe this could be fixed by for instance creating a hook `useLanguages` and doing this pruning there.

PR #125, opened 2022-03-08
Document release 1.0.0-rc3

Jokinen:
### Fixed

- Configured production environment to work from https://liikunta.hel.fi


nikomakela (changes requested):
I would call the next version 1.0.0 instead of release candidate 3. I used the RC when I installed Liikunta with first versions of Unified-Search

PR #129, opened 2022-04-07
LIIKUNTA-275: Fonts

melniiv:
## Description
-Fixed font sized and hline-heights for collection cards

### Closes

**[DEV-XXX](https://helsinkisolutionoffice.atlassian.net/browse/DEV-XXX):**

### DIFF
--- a/src/common/components/card/CollectionCard.tsx
+++ b/src/common/components/card/CollectionCard.tsx
@@ -9,7 +9,7 @@ function CollectionCard({ id, title, infoLines, keywords, href, image }: Item) {
     <Card id={id}>
       <Card.Content>
         <Card.Title variant="h3" title={title} href={href} />
-        <Card.InfoLines as="p" variant="body-l" infoLines={infoLines} />
+        <Card.InfoLines variant="body-l" infoLines={infoLines} />
       </Card.Content>
       <Card.Cta />
       <Card.Keywords keywords={keywords} className={styles.keywords} />
--- a/src/common/components/card/CollectionCard.tsx
+++ b/src/common/components/card/CollectionCard.tsx
@@ -9,7 +9,7 @@ function CollectionCard({ id, title, infoLines, keywords, href, image }: Item) {
     <Card id={id}>
       <Card.Content>
         <Card.Title variant="h3" title={title} href={href} />
-        <Card.InfoLines infoLines={infoLines} />
+        <Card.InfoLines as="p" variant="body-l" infoLines={infoLines} />
       </Card.Content>
       <Card.Cta />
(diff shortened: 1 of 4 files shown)


src/common/components/card/CollectionCard.tsx
```diff
@@ -9,7 +9,7 @@ function CollectionCard({ id, title, infoLines, keywords, href, image }: Item) {
     <Card id={id}>
       <Card.Content>
         <Card.Title variant="h3" title={title} href={href} />
-        <Card.InfoLines infoLines={infoLines} />
+        <Card.InfoLines as="p" variant="body-l" infoLines={infoLines} />
```
Jokinen, line 12: Variant `body-l` should be mapped to the `p` element by default:

https://github.com/City-of-Helsinki/liikunta-helsinki/blob/main/src/common/components/text/Text.tsx#L36

src/common/components/card/LargeCollectionCard.tsx
```diff
@@ -16,7 +16,7 @@ function LargeCollectionCard({
     <Card id={id} className={styles.largeCollectionCard}>
       <Card.Content className={styles.text}>
         <Card.Title as="h3" variant="h2" title={title} href={href} />
-        <Card.InfoLines as="p" variant="body-xl" infoLines={infoLines} />
+        <Card.InfoLines as="p" variant="body-l" infoLines={infoLines} />
```
Jokinen, line 19: Variant `body-l` should be mapped to the `p` element by default:

https://github.com/City-of-Helsinki/liikunta-helsinki/blob/main/src/common/components/text/Text.tsx#L36

src/pages/collections/[slug].tsx
```diff
@@ -33,6 +33,8 @@ type CollectionItemListProps = {
   noPagination?: boolean;
 };
 
+const TextBodyL = (props) => <Text {...props} variant="body-l" />;
```
Comment (Jokinen) [inline src/pages/collections/[slug].tsx:36]: I think props might be `any` here. I think we could consider whether we want to give this a type. For instance with `React.HTMLAttributes<HTMLParagraphElement>`, which should be close to the props a `p` element can receive.

The result is a bit uglier declaration due to slight differences between the attributes of a `p` element and the props of a `Text`.

```tsx
const TextBodyL = ({
  children,
  ...props
}: React.HTMLAttributes<HTMLParagraphElement>) =>
  children ? (
    <Text {...props} variant="body-l">
      {children}
    </Text>
  ) : null;
```

Maybe we could align these APIs by making children optional in the `Text` component?

Your call.

Jokinen (approved):
Looks good, good job 👍

Minor comments on technical details.

PR #130, opened 2022-04-09
HL-274: Large card line limit 3

melniiv:
## Description
Css fix for the frontpage large card to limit the number of lines to show

### Closes

**[DEV-XXX](https://helsinkisolutionoffice.atlassian.net/browse/DEV-XXX):**

## Screenshots
image

### DIFF
--- a/src/common/components/card/Card.tsx
+++ b/src/common/components/card/Card.tsx
@@ -76,6 +76,7 @@ type InfoLineObject = {
 type CardInfoLinesProps = Partial<React.ComponentProps<typeof Text>> & {
   infoLines: (InfoLineObject | string)[];
   className?: string;
+  clamp?: boolean;
 };
 
 function CardInfoLines({
@@ -130,6 +131,10 @@ function CardInfoLines({
       })}
     </div>
   );
+}
+
+function ClampedCardInfoLines(props) {
+  return <CardInfoLines {...props} className={styles.clamp} />;
 }
 
 type CardContentChildrenProps =
@@ -344,6 +349,7 @@ function Card({ children, id, className }: CardProps) {
 Card.Title = CardTitle;
 Card.Pre = CardPre;
 Card.InfoLines = CardInfoLines;
+Card.ClampedCardInfoLines = ClampedCardInfoLines;
 Card.Content = CardContent;
 Card.Cta = CardCta;
 Card.CtaButton = CardCtaButton;
--- a/src/common/components/card/LargeCollectionCard.tsx
+++ b/src/common/components/card/LargeCollectionCard.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 
 import { Item } from "../../../types";
+import formatToSingleTag from "../../utils/formatToSingleTag";
 import Card from "./Card";
 import styles from "./largeCollectionCard.module.scss";
 
@@ -16,7 +17,10 @@ function LargeCollectionCard({
     <Card id={id} className={styles.largeCollectionCard}>
       <Card.Content className={styles.text}>
         <Card.Title as="h3" variant="h2" title={title} href={href} />
-        <Card.InfoLines variant="body-l" infoLines={infoLines} />
+        <Card.ClampedCardInfoLines
+          variant="body-l"
+          infoLines={formatToSingleTag(infoLines)}
+        />
       </Card.Content>
       <Card.Cta className={styles.cta} />
       <Card.Keywords keywords={keywords} className={styles.keywords} />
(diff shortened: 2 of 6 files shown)


src/common/components/card/card.module.scss
```diff
@@ -53,6 +53,14 @@
     outline: 2px solid $color-black;
   }
 
+  & .clamp {
+    display: -webkit-box !important;
+    -webkit-line-clamp: 3;
+    -webkit-box-orient: vertical;
```
Jokinen, line 59: It seems like these webkit specific flavours do not work in Firefox even though Firefox claims they should:

Screenshot 2022-04-11 at 7 45 18

Safari also works poorly:

Screenshot 2022-04-11 at 7 58 07

Perhaps it's the combination of rules that's the issue? Seems like this approach is a bit brittle in some of the major browsers. Perhaps we should consider alternatives like line-height/max-height?

Jokinen (changes requested):
A few comments, otherwise good job 

In the example card, the text ends in an awkward way where ellipsis is preceded by a period and a space:

Screenshot 2022-04-11 at 8 03 05

Is there some way we could avoid this? I couldn't come up with anything we could do about it.

src/common/components/card/card.module.scss
melniiv: interesting, should be supported. Yes, there are other ways with line-height, i ll try

src/common/components/card/Card.tsx
```diff
@@ -132,6 +132,10 @@ function CardInfoLines({
   );
 }
 
+function ClampedCardInfoLines(props) {
+  return <CardInfoLines {...props} className={styles.clamp} />;
+}
```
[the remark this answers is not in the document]
Jokinen: I would have stored this flavour in a separate file as separate module.

src/common/components/card/LargeCollectionCard.tsx
```diff
@@ -16,7 +17,10 @@ function LargeCollectionCard({
     <Card id={id} className={styles.largeCollectionCard}>
       <Card.Content className={styles.text}>
         <Card.Title as="h3" variant="h2" title={title} href={href} />
-        <Card.InfoLines variant="body-l" infoLines={infoLines} />
+        <Card.ClampedCardInfoLines
+          variant="body-l"
+          infoLines={formatToSingleTag(infoLines)}
```
Jokinen, line 22: With this implementation, wouldn't the resulting output be something like `[<undefined>lines</undefined>]`?

src/common/utils/formatToSingleTag.ts
```diff
@@ -0,0 +1,19 @@
...
+  as?: SupportedTags
+): (string | ItemInfoLineObject)[] {
+  if (typeof text[0] === "string") {
+    return [
+      `<${as}>${text[0]
+        .replace(/<[^>]+>/g, " ")
+        .replace(/\s+/g, " ")
+        .trim()}</${as}>`,
+    ];
+  } else {
```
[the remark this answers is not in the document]
melniiv: yes, basically it was for future use if it supports not only string. but i can remove for right now

src/common/utils/formatToSingleTag.ts
```diff
@@ -0,0 +1,19 @@
+import { ItemInfoLineObject } from "../../types";
+
+type SupportedTags = "p";
+
+export default function formatToSingleTag(
+  text: (string | ItemInfoLineObject)[],
```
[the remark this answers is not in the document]
melniiv: it was my first implementation

src/common/components/card/LargeCollectionCard.tsx
```diff
@@ -16,7 +17,10 @@ function LargeCollectionCard({
     <Card id={id} className={styles.largeCollectionCard}>
       <Card.Content className={styles.text}>
         <Card.Title as="h3" variant="h2" title={title} href={href} />
-        <Card.InfoLines variant="body-l" infoLines={infoLines} />
+        <Card.ClampedCardInfoLines
+          variant="body-l"
+          infoLines={[formatToSingleTag(infoLines[0], "p")]}
```
Jokinen, line 22: On this line we are taking the first member of the array. I wonder if that's the best API. I wonder if it'd be clearer for the dev using this component if we looped over all the lines so that we don't accidentally lose any of them.

E.g. "pseudo code"

```ts
lines.map((line, i) => i === 0 ? formatToSingleTag("text" in line ? line.text : line) : line)
```

You aded `p` as the default value for the second parameter, so you no longer need to declare it here.

PR #133, opened 2022-04-12
LIIKUNTA-285 | Fix external link labels

Jokinen:
## Description

Previously we added external link labels by including content in the tag that was readable by a screen reader and told that the link will open in a new window.

In this PR I change that implementation to one where the accessible name of the link is described with `aria-label`. We couldn't get details about the exact usecase where this is vital, but I assume that it has something to do with more complex navigation patterns.

I used HDS' `Link` component to achieve this behaviour. I wanted to use HDS' link in order to align with HDS visually and to avoid re-implementing existing behaviour.

### Closes

**[LIIKUNTA-285](https://helsinkisolutionoffice.atlassian.net/browse/LIIKUNTA-285):**


src/common/components/link/Link.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+    return (
+      // Apply ref directly to HDSLink once it supports it
+      <span ref={ref}>
+        <HDSLink
+          {...delegatedProps}
+          href={href}
+          openInNewTab={isOpenInNewTab}
+          external={isExternal}
+          size="M"
```
melniiv, line 27: should size be configurable as well with some default?

melniiv (approved):

src/common/components/link/Link.tsx
Jokinen: I attempted to use this size property to adjust font sizes for some components. The issue is that in HDS the icon scales more than the text, which is in conflict with the design that has been made for this service.

Screenshot 2022-04-13 at 7 59 41

My thinking was that for now, using the prop would most likely lead to a result that conflicts with the design. And in that way it doesn't make sense to expose the prop. It should be safe to expose later on as well if the need arises.

Jokinen: I've exposed the prop for now 👍

PR #139, opened 2022-04-19
LIIKUNTA-185 | Redirect to error page on apollo client network errors

Jokinen:
## Description

1. Add a generic error page
2. Redirect user to generic error page when  an apollo client faces a network error

I tested this by blocking requests into unified search:

If the error happens server side while generating a static page, nothing is done. The broken page is served and the user is navigated into the `/error` view with `push`.

It'd be possible to redirect instead of serving a broken page server side, but I'm not sure whether it's better or worse UX, nor whether it's better or worse in a technical sense. In this instance we wanted to fix behaviour on the search page when US is unreachable. US related queries should be client side only.

Screenshot 2022-04-19 at 15 32 39

### Closes

**[LIIKUNTA-185](https://helsinkisolutionoffice.atlassian.net/browse/LIIKUNTA-185)**


melniiv (approved):

Jokinen:
Waiting for translations before merging.

Jokinen:
AFAIK translations don't yet exits. My tenure at City of Helsinki is ending.

For the future contributor: the only thing missing should be the translations.

